PR #12, opened 2022-10-31
User

reganq:
I created the user class. Importantly here is the password check which can be used for signing in, and also the getter and setter methods for the favourites and posts list


src/main/java/entities/User.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class User;
+    ArrayList<Post> favourites;
+        ArrayList<Post> posts;
+        boolean isAdmin;
+        String userName;
+        String password;
+
+public User(String name, String pass, boolean admin) {
+        this.favourites = ArrayList();
```
JohannLeidenfrost, line 11: this.favourites = new ArrayList();

src/main/java/entities/User.java
```diff
@@ -0,0 +1,65 @@
+package entities;
+
+public class User;
```
JohannLeidenfrost, line 3: Syntax issue: public class User{...}

src/main/java/entities/User.java
sushimon, line 11: Need to initialize an ArrayList with the new keyword for favourites and posts attributes.

src/main/java/entities/User.java
```diff
@@ -0,0 +1,65 @@
...
+    ArrayList<Post> favourites;
```
JohannLeidenfrost, line 4: add access modifiers for variables

src/main/java/entities/User.java
```diff
@@ -0,0 +1,65 @@
...
+public void addPost(Post toAdd){
+        /* add toAdd to this.posts */
+        this.posts.add(toAdd);
+        }
+
+public boolean removePost(Post toRemove){
+        /* Search for toRemove in this.posts and if it is the remove and return true,
+         * otherwise return false */
+        for (int i; i < this.posts.size(); i ++){
+        if (this.posts.get(i).equals(toRemove)){
```
JohannLeidenfrost, line 60: override .equals method for post class

izumidayo (changes requested):
1. line 31, for loop needs i=0
2. line 61, for loop needs i=0

src/main/java/entities/User.java
```diff
@@ -0,0 +1,65 @@
...
+public void addFavourite(Post toAdd){
+        /* Add Post to this.favourites */
+        this.favourites.add(Post);
+        }
+
+public boolean removeFavourite(Post toRemove){
+        /* Search for Post in this.favourites, remove the first instance found.
+         * Return true if found and removed and return false otherwise */
+        for (int i; i < this.favourites.size(); i ++){
+        if (this.favourites.get(i).equals(toRemove)){
```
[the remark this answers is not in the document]
reganq: Post equality has been overridden

src/main/java/entities/User.java
```diff
@@ -0,0 +1,133 @@
...
+
+    // id is the unique identifier of User, equal to number of posts at time of creation
+    private int id;
+
+    /*
+    * Initializes an instance of User
+    *
+    * @param admin whether the user will have admin priviledges
+    * */
+    public User(boolean admin) {
```
JohannLeidenfrost, line 24: You could add a parameter for the number of users in the system since we do not want this number to restart at 0 every time the program is re-run. So we would probably save the number of users in the database and, when creating a new one, get that data from the DB and feed it as a parameter into this constructor and then assign the new user the id = number of users + 1 (and then increment number of users by 1). What do you think? This is the way I implemented the posts.

src/main/java/entities/User.java
```diff
@@ -0,0 +1,133 @@
...
+    // ArrayList of Posts User has made
+    private ArrayList<Post> posts;
+
+    // boolean relating if User has admin privileges
+    private boolean isAdmin;
+
+    // Number of Users created
+    public static int numUsers = 0;
+
+    // id is the unique identifier of User, equal to number of posts at time of creation
```
JohannLeidenfrost, line 16: Equal to the number of posts or the number of users at the time of creation?

JohannLeidenfrost (dismissed):
Check out the comment I wrote about the user id and lmk what your opinion is. Looks good otherwise!

src/main/java/entities/User.java
reganq: I meant to use user, my bad!

src/main/java/entities/User.java
reganq: Looked into this and it seems that class variables don't persist with databases, so I'll switch mine over to your implementation

JohannLeidenfrost (dismissed):

src/main/java/entities/User.java
```diff
@@ -0,0 +1,133 @@
...
+    private int id;
+
```
lisan-y, line 18: attributes for email, password

src/main/test/UserTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void testConstructerFavourites() {
+        User user = new User(false, 0);
+        boolean actual = user.getFavourites;
+        boolean expected = true;
+
```
lisan-y, line 21: user.getFavourites should return a list. get.favourites needs ()

reganq: Right good catch

src/main/java/entities/User.java
reganq: Adding these attributes, and adding them to constructor and getter methods.

izumidayo (commented):
Just left a comment on testUserId

src/main/test/UserTest.java
```diff
@@ -0,0 +1,53 @@
...
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void testUserId() {
+        User user = new User(false, 0);
+        int actual = user.id;
+        int expectd = 1;
```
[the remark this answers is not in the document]
reganq: I don't think that would work but as long as the test isn't interacting with the database and no other users have been defined in the test space, we shouldn't have a problem.

src/main/java/entities/Postable.java
```diff
@@ -1,4 +1,4 @@
-package entities;
+package java.entities;
```
sushimon, line 1: Don't think this is needed. Every entity is in the same directory so we can just use "package entities"

src/main/java/entities/User.java
```diff
@@ -0,0 +1,170 @@
+package java.entities;
```
sushimon, line 1: See comment in Postable

src/main/java/entities/User.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.ArrayList;
+
+public class User {
+    // ArrayList of User's favourited posts
+    private final ArrayList<Post> favourites;
```
sushimon, line 6: Change data type to List to keep it more general

src/main/java/entities/User.java
```diff
@@ -0,0 +1,170 @@
...
+
+    // ArrayList of Posts User has made
+    private final ArrayList<Post> posts;
```
sushimon, line 9: See comment on line 6

src/main/java/entities/User.java
```diff
@@ -0,0 +1,170 @@
...
+    /*
+    * Initializes an instance of User
+    *
+    * @param admin whether the user will have admin priviledges
+    * */
+    public User(boolean admin, int numUsersCreated, String email, String password) {
+        this.favourites = new ArrayList<Post>();
+        this.posts = new ArrayList<Post>();
+        this.isAdmin = admin;
+        User.numUsers = numUsersCreated;
```
sushimon, line 35: Since it's static and we increment it each time a user is created, there's no need to have it as a parameter or re-assign numUsers. Just having the increment should be fine.

src/main/java/entities/User.java
```diff
@@ -0,0 +1,170 @@
...
+        this.posts.add(toAdd);
+    }
+
+    /*
+    * Check if a Post is in User's posts and remove it if so
+    *
+    * @param toRemove post to be removed
+    * @returns a boolean representing success or failure
+    * */
+    public boolean removePost(Post toRemove) {
```
sushimon, line 104: Maybe rename so we don't mistake it as part of the removePost use case.

src/main/java/entities/User.java
```diff
@@ -0,0 +1,170 @@
...
+        this.favourites.add(toAdd);
+    }
+
+    /*
+    * Check if a Post is in User's favourites and remove it if so
+    *
+    * @param toRemove Post to be removed
+    * @return boolean representing success or failure
+    * */
+    public boolean removeFavourite(Post toRemove) {
```
sushimon, line 68: Same with comment on line 104, make name more descriptive of what it actually does without conflicting with another use case, etc.

src/main/java/entities/User.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /*
+    * Return if User is equal to o
+    *
+    * @param o The Object to be compared to
+    * @return boolean representing whether User is equal to o
+    * */
+    @Override
+    public boolean equals (Object o) {
+        if (this == o) {
```
sushimon, line 133: Seems like a redundant check since it's covered in the final check on line 139 onwards

src/main/java/entities/User.java
```diff
@@ -0,0 +1,170 @@
...
+            return true;
+        } else if (o == null) {
+            return false;
+        } else if (!(o instanceof User)) {
+            return false;
+        }  else {
+            return (this.id == ((User) o).id);
```
sushimon, line 140: Nit: do the typecasting on another line to make code more readable

sushimon (changes requested):
Left some comments on possible changes but the rest lgtm!

src/main/java/entities/User.java
reganq: We discussed this and realized that the class variable doesn't persist in the database, so we need this feature to keep a proper id system.

src/main/java/entities/Postable.java
reganq: This was needed for my intellij to work but I can change it for the repo.

src/main/java/entities/User.java
reganq: changed

src/main/java/entities/User.java
reganq: not sure how to do this so I'll just leave it for now

src/main/java/entities/User.java
reganq: removed

src/main/java/entities/User.java
izumidayo: Is there some convention that helps people label methods so you can easily tell whether they are part of entities or use cases? Otherwise I can't really think of a good way to rename it. Maybe "removePostE"?

izumidayo (approved):

src/main/java/entities/User.java
sushimon: Have something like `User user = (User) o` before line 139 and then just compare this.id and user.id

sushimon (approved):
lgtm

BerryRahul (approved):

JohannLeidenfrost (approved):
Looks good!

PR #13, opened 2022-10-31
Create Tag entity and unit tests.

sushimon:
Created the Tag entity and a unit test.

Will complete test suite once Post gets pushed.

### DIFF
--- a/src/main/java/entities/Tag.java
+++ b/src/main/java/entities/Tag.java
@@ -1,0 +1,37 @@
+package entities;
+
+import java.util.ArrayList;
+
+public class Tag {
+    // The name of the tag
+    final String name;
+
+    // Array of posts that are under this tag
+    final ArrayList<Post> taggedPosts;
+
+    /**
+     * Initializes an instance of Tag.
+     *
+     * @param name The name of this tag.
+     */
+    public Tag(String name) {
+        this.name = name;
+        this.taggedPosts = new ArrayList<>();
+    }
+
+    /**
+     * Add a post that has this tag.
+     *
+     * @param post The post that has this name as a tag.
+     */
+    public void addTag(Post post) {
+        taggedPosts.add(post);
+    }
+
+    /**
+     * Returns the name of this tag.
+     */
+    public String getName() {
+        return this.name;
+    }
+}
--- a/src/main/java/entities/Tag.java
+++ b/src/main/java/entities/Tag.java
@@ -29,6 +29,11 @@ public class Tag {
     }
 
     /**
+     * Returns whether the post that has this tag was successfully removed.
+     */
+    public boolean removeTag(Post post) { return taggedPosts.remove(post); }
+
+    /**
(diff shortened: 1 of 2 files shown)


reganq (changes requested):
Should add a removeTag option to remove posts that have been deleted.

src/main/java/entities/Tag.java
```diff
@@ -0,0 +1,37 @@
...
+    final ArrayList<Post> taggedPosts;
+
+    /**
+     * Initializes an instance of Tag.
+     *
+     * @param name The name of this tag.
+     */
+    public Tag(String name) {
+        this.name = name;
+        this.taggedPosts = new ArrayList<>();
```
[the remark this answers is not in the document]
sushimon: Yes, I think we agreed that comments wouldn't have tags

reganq (approved):

src/main/java/entities/Tag.java
```diff
@@ -0,0 +1,42 @@
+package entities;
+
+import java.util.ArrayList;
+
+public class Tag {
+    // The name of the tag
+    final String name;
+
+    // Array of posts that are under this tag
+    final ArrayList<Post> taggedPosts;
```
JohannLeidenfrost, line 10: Could you make this a List instead of ArrayList just so we can keep is as general as possible.

src/main/java/entities/Tag.java
```diff
@@ -0,0 +1,42 @@
...
+     * @param post The post that has this name as a tag.
+     */
+    public void addTag(Post post) {
+        taggedPosts.add(post);
+    }
+
+    /**
+     * Returns whether the post that has this tag was successfully removed.
+     */
+    public boolean removeTag(Post post) { return taggedPosts.remove(post); }
```
JohannLeidenfrost, line 34: Consider renaming the method to something like removePost since we are removing a post, not a tag. Also, I think it is necessary to iterate through the taggedPosts, use the .equals method for Posts (overridden in Post entity; it compares posts based on their ID), and then remove the Post if its ID is equal to the ID of the Post that is to be deleted. This way there will be no issue with whether two posts that have the same id actually refer to the same object in memory.

JohannLeidenfrost (changes requested):
Please check the comments I left, especially regarding removePost.

src/main/java/entities/Tag.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * Add a post that has this tag.
+     *
+     * @param post The post that has this name as a tag.
+     */
+    public void addTag(Post post) {
+        taggedPosts.add(post);
+    }
```
lisan-y, line 29: should we check if the post is already in the tags list (is it possible we could fuck that up somewhere in the code?) and throw an exception?

JohannLeidenfrost (approved):

izumidayo (approved):

PR #14, opened 2022-10-31
Created Comment Entity and addComment method in Postable

izumidayo:
- Updated Postable to implement addComment method
- Created new class Comment which extends Postable

### DIFF
--- a/src/main/java/entities/Comment.java
+++ b/src/main/java/entities/Comment.java
@@ -2,9 +2,17 @@ package entities;
 
 public class Comment extends Postable{
     // Class Comment extends abstract class Postable and inherits all of its instance
-    // (user, body, replies) and addComment method (which has already been implemented)
+    // attributes (user, body, replies) and addComment method
+    // (which has already been implemented)
     public Comment(User commenter, String comment){
         super.user = commenter;
         super.body = comment;
+    }
+    public String getBody(){
+        return super.body;
+    }
+
+    public User getUser(){
+        return super.user;
     }
 }
--- a/src/main/java/entities/Comment.java
+++ b/src/main/java/entities/Comment.java
@@ -1,0 +1,10 @@
+package entities;
+
+public class Comment extends Postable{
+    // Class Comment extends abstract class Postable and inherits all of its instance
+    // and addComment method (which has already been implemented)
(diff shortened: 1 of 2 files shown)


reganq (dismissed):
For Comment, I think we should add some getter methods, for user as Lisa said but also for body. I think it would be useful when we are implementing the view

JohannLeidenfrost (approved):

JohannLeidenfrost (approved):

reganq (dismissed):
Everything looks good!

sushimon (dismissed):
lgtm! just remember add some test cases once all the entities have been merged

lisan-y (approved):

src/main/java/entities/Postable.java
```diff
@@ -5,6 +5,12 @@
 abstract class Postable{
     User user;
     String body;
-    List<Comment> replies;
+
+    List<Postable> replies;
```
JohannLeidenfrost, line 9: Can you add a getter for replies?

src/main/java/entities/Postable.java
```diff
@@ -5,6 +5,12 @@
 abstract class Postable{
     User user;
```
JohannLeidenfrost, line 6: Can you add access modifiers for the instance variables?

src/main/java/entities/Comment.java
```diff
@@ -8,14 +9,17 @@
 
 
 public class Comment extends Postable{
-
+    // Class Comment extends abstract class Postable and inherits all of its instance
+    // attributes (user, body, replies) and addComment method
+    // (which has already been implemented)
     public static Date creationDate;
```
JohannLeidenfrost, line 15: Since we are adding comments as replies to the Post, I think we will have to make sure to not accidentally add the same comment twice. Could you add a system of IDs to the comments (similar to what we did for User and Post) and a .equals method that checks for ID equality? In that case, we could probably make the ID a static variable for Portable so that Comment and Post both have access to it and can use the same .equals method (unless we want to keep track of the number of Posts and Comments separately.

reganq (approved):

sushimon (approved):

PR #15, opened 2022-10-31
Added Post entity as well as incomplete test cases for Post.

JohannLeidenfrost:


src/main/java/entities/post.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * The Post class contains all the necessary information of a post: its ID, title,
+ * description,tags,collaborators, deadline, and users who favourited it.
+ */
+public class Post extends Postable{
+    public static int numPostsCreated = 0;
+    /**The id is equal to the number of posts created.*/
+    private int id;
+    private Timer timer;
```
reganq, line 13: I think we agreed to ditch the Timer object, the deadline is going to be controlled by calls to a controller in the main loop.

src/main/java/entities/post.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * @param poster the User who is posting this Post. This is an attribute of the superclass Postable.
+     * @param timer an object that keeps track of time and deletes the Post when the deadline arrives.
+     * @param title this Post's title.
+     * @param mainDesc this Post's description.
+     * @param tags the tags this Post is added to.
+     * @param collaborators a description of the type of the collaborators the User who posted this Posts is looking for.
+     * @param deadline the date on which the Post is to be deleted from the database.
+     * @param replies the comments that are added to this post. This is an attribute of the superclass Postable.
+     * @param numPostsCreated1 the number of posts created so far.
```
reganq, line 30: Pedantic I know, but small typo in Line 30 numPostsCreated

src/main/java/entities/post.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    public Post(User poster, Timer timer, String title, String mainDesc, List<String> tags, String collaborators,
+                Date deadline, List<Comment> replies, int numPostsCreated1){
+        super();
```
reganq, line 34: Should the call to super() contain the parameters it accepts?

reganq (commented):
It looks good besides my minor comments so I'm going to go ahead and approve. Just don't forget about them.

reganq (approved):

src/main/java/entities/post.java
JohannLeidenfrost: Because numPostsCreated is a static variable, I thought it would be better if I make the parameter a different name than numPostsCreated since I cannot use "this." with a static var.

src/main/java/entities/post.java
JohannLeidenfrost: Postable only has the default constructor so the explicit call to super is probably not necessary!

PR #16, opened 2022-11-01
4 feature 9 make a new post: Added Post entity and incomplete tests and removed timer object

JohannLeidenfrost:


reganq (dismissed):
Looks Good! Although the PostBody still needs to be implemented though. Approving to get the Post into the main repo

src/main/java/entities/post.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public class Post extends Postable{
+    public static int numPostsCreated = 0;
+    /**The id is equal to the number of posts created.*/
+    private int id;
+    /**This is a list of the users who favourited this post.*/
+    private List<User> favouritedUsers;
+    private String title;
+    private List<String> tags;
+    private String collaborators;
```
izumidayo, line 17: Should collaborators be a list of users?

izumidayo (commented):
Question about instance attribute collaborators

src/main/java/entities/post.java
reganq: I think collaborators is a presented attribute, sort of like qualifications, etc.

reganq (dismissed):
I think it would be helpful to have a getter method for the date, but I can add that when I work on the timer feature. Looks good otherwise.

src/main/java/entities/post.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param title this Post's title.
+     * @param mainDesc this Post's description.
+     * @param tags the tags this Post is added to.
+     * @param collaborators a description of the type of the collaborators the User who posted this Posts is looking for.
+     * @param deadline the date on which the Post is to be deleted from the database.
+     * @param replies the comments that are added to this post. This is an attribute of the superclass Postable.
+     * @param numPostsCreated1 the number of posts created so far.
+     */
+    public Post(User poster, String title, String mainDesc, List<String> tags, String collaborators,
+                Date deadline, List<Comment> replies, int numPostsCreated1){
```
sushimon, line 31: Parameter list might be too long, code smell

src/main/java/entities/post.java
```diff
@@ -0,0 +1,130 @@
...
+        super.user = poster;
+        super.body = mainDesc;
+        super.replies = replies;
+        numPostsCreated = numPostsCreated1;
```
sushimon, line 35: If numPostsCreated is static I don't think we need to assign it here or have it in the parameter list of the constructor. The increment at the end of the constructor should be fine.

src/main/java/entities/post.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Searches for userToRemove in the list of Users who favourited the Post, removes userToRemove if
+     * that user is there.
+     * @param userToRemove the User to be removed.
+     * @return true if the user has been removed; false otherwise.
+     */
+    public boolean removeFavouritedUser(User userToRemove){
+        for(int i = 0; i < favouritedUsers.size(); i++){
```
sushimon, line 93: Nitpick: looks like you could replace with a for-each loop instead (same function)

src/main/java/entities/post.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Checks whether two posts are equal based on whether their Post id's are equal.
+     * @param o the Post to be compared to.
+     * @return a boolean that reflects whether the two Posts' id's are equal.
+     */
+    @Override
+    public boolean equals(Object o){
+        if(this == o){
```
sushimon, line 117: Correct me if I'm wrong but isn't this a redundant check? I think this is covered by the check at the end of the method.

sushimon (changes requested):
Left some comments on places that might need changes but the rest lgtm!

BerryRahul (approved):

src/main/java/entities/post.java
JohannLeidenfrost: I think you're right! It's just that they implemented it this way in the class notes so I copied it; I am not exactly sure about the reason though.

src/main/java/entities/post.java
JohannLeidenfrost: Yes, we need it in the constructor because we want to ensure data persistence so even when the program is restarted, we get numPostsCreated from the database and pass it through the constructor so we don't start at 0 again.

src/main/java/entities/post.java
JohannLeidenfrost: You're right but I think the fix would be to make a class that takes these parameters as instance variables. That, however, would be another code smell (data class) so I thought I would leave it this way for now.

src/main/java/entities/post.java
lisan-y: do we need collaborators?

sushimon (dismissed):

reganq (dismissed):
Everything here looks good to me.

PR #18, opened 2022-11-07
Changed file structure and added Post and PostTests.

JohannLeidenfrost:


sushimon (approved):
lgtm

src/main/java/entities/Post.java
```diff
@@ -0,0 +1,168 @@
...
+     * @return this post's deadline.
+     */
+    public Date getDeadline(){
+        return this.deadline;
+    }
+
+    /**
+     * @return the replies made to this Post.
+     */
+    public List<Postable> getReplies(){
```
sushimon, line 81: Might need to change return type to List<Comment>, the autograding says it can't convert List<Comment> to List<Postable>

sushimon (changes requested):

src/main/java/entities/Post.java
JohannLeidenfrost: Can you clarify? I thought the return type is List already

sushimon: Sorry, I don't know why it ignored the stuff inside the chevrons. It should say change List\<Postable\> to List\<Comment\>

JohannLeidenfrost: I think the problem is that, currently, the type for replies is List<Comment> but when the new PR for comment is approved, it will be List<Postable> and then it will work.

sushimon: Ah yes, just checked the other PR. Everything is fine then

sushimon (approved):

reganq (approved):
Looks good

izumidayo (approved):

PR #19, opened 2022-11-07
Refactor code & fix errors

sushimon:
CurrentUser File:
- getCurrentUser's method body didn't call the correct method
- getIsAdmin tried accessing private field

User File:
- Return types of some methods were using raw types
- Fixed documentation of some methods


JohannLeidenfrost (approved):
Looks good!

reganq (approved):
Looks good, thanks for the catch on the user!

izumidayo (approved):
very very good catch here, looks good to me!

PR #22, opened 2022-11-08
Update unit tests in TagTest and fix misc. code

sushimon:
Changes:
- Updated `build.gradle` file to fix test errors (as per that one Piazza post)
- Changed Postable.addComment to use `ArrayList.add` as there is no `ArrayList.append`
- Updated unit tests in TagTest to test each method

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,12 +8,11 @@ repositories {
 
 dependencies {
     implementation 'junit:junit:4.13.1'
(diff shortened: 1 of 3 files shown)


izumidayo (approved):
Thanks for catching the append / add issue & for fixing the autotest issues as well. It looks good to me!

BerryRahul (approved):

reganq (approved):
looks good!

PR #26, opened 2022-11-10
Changes to Postable and Comment entities

izumidayo:
- created getReplies and getCreationDate methods on Postable
- added doc comments on class Comment and its methods
- overrode equals method for Comment
- added id attribute in Comment (implemented the same way as User)


sushimon (changes requested):
check the comments i left. everything else lgtm!

src/main/java/entities/Postable.java
```diff
@@ -1,15 +1,34 @@
 package entities;
 
 import java.util.List;
+import java.util.Date;
 
 abstract class Postable{
```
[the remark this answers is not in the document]
izumidayo: done

src/main/java/entities/Comment.java
```diff
@@ -7,30 +7,62 @@
 
 import java.security.PublicKey;
```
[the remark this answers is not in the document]
izumidayo: removed!

sushimon (dismissed):
lgtm!

src/main/java/entities/Comment.java
```diff
@@ -5,32 +5,62 @@
...
+        super.creationDate = new Date();
+        Comment.numComments = numCommentsCreated;
+        Comment.numComments ++;
+        this.id = Comment.numComments;
     }
+
+    /**
+     * Returns the body (or text content) of this instance of Comment.
+     */
     public String getBody(){
```
reganq, line 39: This method seems like something that the abstract class should already have implemented. We could change it to just be a method in super and then Comment and Post would already have these implemented

src/main/java/entities/Comment.java
```diff
@@ -5,32 +5,62 @@
...
         return super.body;
     }
 
+    /**
+     * Returns the User that created this Comment.
+     */
     public User getUser(){
```
reganq, line 46: Similar argument to line 24, in that this could also be a method in Postable.

reganq (changes requested):
I think we should add getter/setter methods for the attributes of Postable, and then let Comment and Post inherit them. Other than that, everything looks good!

src/main/java/entities/Comment.java
izumidayo: moved!

BerryRahul (approved):

src/main/java/entities/Postable.java
```diff
@@ -1,15 +1,53 @@
...
+
+    // the User who created the Postable
     User user;
+    // the main text content of the Postable
     String body;
-
+    // a list of Postable objects created as a reply to this Postable
     List<Postable> replies;
+    // the date this Postable was created
+    Date creationDate;
```
sushimon, line 17: Maybe something to talk about with Johann. In the Post entity, he uses LocalDate to store the date of the deadline (different attribute but same idea) so maybe ask him just to be sure. 

Just as a note if you decide to change everything to LocalDate, `LocalDate.now()` returns the current date.

src/main/java/entities/Postable.java
```diff
@@ -1,15 +1,53 @@
...
+
+    /**
+     * Returns the list of replies to a Postable object.
+     */
+    public List<Postable> getReplies(){return this.replies;}
+
+    /**
+     * Returns the date of which this Postable was created.
+     */
+    public Date getCreationDate(){return this.creationDate;}
```
sushimon, line 36: Might need to change return type and docstring after discussing with Johann about whether to use `Date` vs `LocalDate`

src/main/java/entities/Postable.java
```diff
@@ -1,15 +1,53 @@
 package entities;
 
 import java.util.List;
+import java.util.Date;
```
sushimon, line 4: Might need to change if you decide to use `LocalDate` instead

src/main/java/entities/Comment.java
```diff
@@ -5,32 +5,48 @@
...
+    public Comment(User commenter, String body, int numCommentsCreated){
         super.user = commenter;
         super.body = body;
         super.replies = new ArrayList<>();
-        creationDate = new Date();
-
-    }
-    public String getBody(){
-        return super.body;
+        super.creationDate = new Date();
```
sushimon, line 30: Refer to other comments, similar idea.

sushimon (changes requested):
Discuss with Johann about whether to use `Date` or `LocalDate` as it'd be weird to use two different types for two similar attributes imo

sushimon (approved):
Going to approve so we can get this merged, but something TODO is to discuss whether to use `Date` or `LocalDate`

PR #27, opened 2022-11-14
My in progress sign up use case. In the description I've put a couple notes/unfinished parts. Pull request is for feedback/help with issues

reganq:
Created all layers of clean architecture for Sign Up use case, with tests for the use case interactor and the viewmodel. Also contains a test file for view, but this was only for opening the window.

Some notes: 1) The tests in SignUpViewModelTest and SignUpTest do not actually run on my device, even though the IDE does not have any problems with them. It is an open issue on Piazza for me, but I've decided to push through so the rest of the code can be checked. 2) The code for changing the view in the event of successful user creation has been put off, as we agreed to setup the main view together. 3) The implementation for the database access is not done, because I haven't looked into it much yet. Gonna try to get my other use case done first. 4) In the view, I have been unable to get the image for the logo to appear.


src/main/java/sign_up/use_case/SignUpInteractor.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Override
+    public void signUp(SignUpRequestModel requestModel) {
+        try {
+            SignUpResponseModel responseModel = this.signUpHelper(requestModel);
+            this.outputBoundary.present(responseModel);
+        }
+        catch (MakeUserException e) {
+            SignUpResponseModel responseModel = new SignUpResponseModel(false, e.getMessage());
+        }
```
BerryRahul, line 31: shouldn't there be a 'this.outputBoundary.present(responseModel) here as well? in order to present the response with the error message

reganq: Yes good catch, thank you!

src/main/java/sign_up/interface_adapters/SignUpViewModel.java
```diff
@@ -0,0 +1,35 @@
+package sign_up.interface_adapters;
+
+import sign_up.use_case.*;
+
+public class SignUpViewModel implements SignUpOutputBoundary {
```
sushimon, line 5: I think you might be able to redesign the UI/interface adapter interactions by implementing the observer pattern by using stuff from the `java.beans` package instead (see [design pattern repo](https://github.com/CSC207-UofT/design-pattern-samples/tree/main/src/observer/withObserverJava9AndLater)).

src/main/java/sign_up/interface_adapters/SignUpViewModel.java
```diff
@@ -0,0 +1,35 @@
...
+    // The view which will be updated following the viewmodel
+    private final View view;
+    private final SignUpInputBoundary inputBoundary;
+
+    public SignUpViewModel(SignUpDataAccessInterface dataAccess, View view) {
+        inputBoundary = new SignUpInteractor(dataAccess, this);
```
sushimon, line 11: Is this acting as both the presenter and the controller? Might be better to split into different classes (probably less likely to violate Clean Architecture). As it is now, I think this might be violating the Single Responsibility Principle.

src/main/java/sign_up/interface_adapters/SignUpViewModel.java
```diff
@@ -0,0 +1,35 @@
+package sign_up.interface_adapters;
+
+import sign_up.use_case.*;
+
+public class SignUpViewModel implements SignUpOutputBoundary {
+    // The view which will be updated following the viewmodel
+    private final View view;
+    private final SignUpInputBoundary inputBoundary;
+
+    public SignUpViewModel(SignUpDataAccessInterface dataAccess, View view) {
```
sushimon, line 10: Can we use the "Gateway" suffix for the data access to keep things consistent?

src/main/java/sign_up/interface_adapters/View.java
```diff
@@ -0,0 +1,6 @@
+package sign_up.interface_adapters;
+
+public interface View {
+    public void update(boolean result, String message);
+    public void view();
```
sushimon, line 5: What is this method for exactly? IIRC this wasn't part of the observer pattern (if this is the design pattern that you're using).

src/main/java/sign_up/use_case/SignUpRequestModel.java
```diff
@@ -0,0 +1,30 @@
+package sign_up.use_case;
+
+public class SignUpRequestModel {
+    // The inputted email for the account
+    public String email;
```
sushimon, line 5: These all should be private to avoid any unintentional reassignments, etc. Make these private fields and add getters.

src/main/java/sign_up/use_case/SignUpInteractor.java
```diff
@@ -0,0 +1,80 @@
...
+
+
+    }
+
+    private SignUpResponseModel signUpHelper(SignUpRequestModel requestModel) throws MakeUserException {
+        if (requestModel.email.isEmpty() || requestModel.password.isEmpty()) {
```
sushimon, line 37: Shouldn't be accessing fields of another class directly imo. Make these attributes private in the corresponding class and add getters and call them here (and elsewhere where the same idea applies).

src/main/java/sign_up/use_case/SignUpResponseModel.java
```diff
@@ -0,0 +1,20 @@
+package sign_up.use_case;
+
+public class SignUpResponseModel {
+    // Whether the attempt to create user was successful
+    public boolean creationSuccess;
```
sushimon, line 5: These all should be private to avoid any unintentional reassignments, etc. Make these private fields and add getters.

src/test/java/sign_up/interface_adapters/SignUpViewModelTest.java
```diff
@@ -0,0 +1,149 @@
+package sign_up.interface_adapters;
+
+import org.junit.Before;
+import org.junit.jupiter.api.Assertions;
```
sushimon, line 4: I think the issues with your test files come from importing from `org.junit.jupiter.api` instead of just `org.junit` (without the `jupiter.api`. For example use `import org.junit.Test` instead.

sushimon, line 4: Can replace with `import static org.junit.Assert.*` so you don't have to do `Assertions.____` every time.

src/test/java/sign_up/use_case/SignUpTest.java
```diff
@@ -0,0 +1,128 @@
+package sign_up.use_case;
+
+import entities.User;
+import org.junit.Before;
+import org.junit.jupiter.api.Assertions;
```
sushimon, line 5: See previous comments in `SignUpViewModelTest.java`.

sushimon (commented):
General Feedback:
- Turn public fields of classes into private. Part of OOP is encapsulation so we shouldn't be letting other classes change/access another class' attributes directly. Use getters and setters where applicable.
- Add more documentation for your classes, methods, etc.
- Avoid the inner folders in `test/java/sign_up`. Just have your tests in `test/java/sign_up`.
- Import from `org.junit` instead of `org.junit.jupiter.api`.
- You could test the functionality of most of your use case in one test if you initialize and connect all the different parts of your use case. You can refer to the tests of the filter post use case.

src/main/java/sign_up/interface_adapters/SignUpViewModel.java
reganq: Okay, I'll look into this. Thanks!

src/main/java/sign_up/interface_adapters/SignUpViewModel.java
reganq: I had the same concern, but I've talked to the professors about this at office hours and they say that it is okay in this case being the use of the MVVM model

src/main/java/sign_up/interface_adapters/SignUpViewModel.java
reganq: For sure I will update that

src/main/java/sign_up/interface_adapters/SignUpViewModel.java
reganq: Looking at the way you've implemented the Controller and Presenter with you VM, is this how MVVM is supposed to work? I can implement in a similar way if needed.

src/test/java/sign_up/interface_adapters/SignUpViewModelTest.java
reganq: omg that was actually it. Thank you sm

reganq: Done, thank you!

src/main/java/sign_up/interface_adapters/View.java
reganq: It was basically to allow me to display the original UI for signing up, but I'll look into rehauling my VM with a controller and presenter, similar to the Filter Posts use case.

src/main/java/sign_up/ui/SignUpView.java
```diff
@@ -0,0 +1,111 @@
...
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.beans.PropertyChangeSupport;
+
+
+public class SignUpView extends JPanel implements ActionListener {
+    private final SignUpController signUpController;
```
BerryRahul, line 16: What Is this Class? I don't see it in this commit. 
Also isn't SignUpUserInputData acting as the Controller?

reganq: This is the panel for taking user input when signing up. Its analagous to the login view. if you look at the sign up view test, you can see what i mean.
Also, I think I just forgot to push my file for the controller. you should be able to see it now.

src/main/java/sign_up/drivers/SignUpDatabaseAccess.java
```diff
@@ -0,0 +1,191 @@
...
+        ArrayList<String[]> numUsers;
+        try {
+            numUsers = readLineByLine(numUsersCreatedFilePath);
+        } catch(Exception e) {
+            System.out.println("Wrong Path");
+            String[] defaultEntry = {"0"};
+            numUsers = (ArrayList<String[]>) Arrays.asList(new String[][]{defaultEntry});
+        }
+
+        return Integer.parseInt(numUsers.get(1)[0]);
```
izumidayo, line 39: I did a quick search online and could not figure out what the parseint method does. can you explain this to me? sorry for the hassle

src/main/java/sign_up/ui/SignUpView.java
```diff
@@ -0,0 +1,144 @@
...
+                    break;
+                case ("email exists"):
+                    JOptionPane.showMessageDialog(errorFrame, "That email already exists. Try signing in!",
+                            "Error", JOptionPane.ERROR_MESSAGE);
+                    break;
+                case ("admin password"):
+                    JOptionPane.showMessageDialog(errorFrame, "The admin password is incorrect. " +
+                                    "If you do not want to be admin, leave it blank.", "Error",
+                            JOptionPane.ERROR_MESSAGE);
+                    break;
```
reganq, line 111: Note to self: add the code to make frame visible

BerryRahul (dismissed):
looks all good to me

BerryRahul (dismissed):
lgtm

izumidayo (dismissed):
SignUpDatabaseAccess might need some adjustment once we have a unified database and testing needs to be updated but from what you have showed me and what I have read over the rest looks good to me! Let's merge to main for now because the fixes I mention above won't disturb the rest of the code.

src/main/java/sign_up/use_case/SignUpDSGateway.java
```diff
@@ -0,0 +1,17 @@
+package sign_up.use_case;
+
+import entities.User;
+
+import java.util.List;
+
+public interface SignUpDSGateway {
```
sushimon, line 7: Rename to DsGateway

src/test/java/entities/PostTest.java
```diff
@@ -114,15 +114,15 @@ public void getDeadline() {
     public void getReply(){
         LocalDate deadline = LocalDate.of(2018, 12, 31);
         Post newPost = new Post(poster, title, mainDesc, tags, collaborators, deadline, 0);
-        Comment comment = new Comment(poster, "Hello!");
+        Comment comment = new Comment(poster, "Hello!", 0);
```
sushimon, line 117: This looks like it's still using the old entities. For example, `Comment` no longer accepts a `User` object as a parameter. I think Johann's PR fixes this so you can remove this file from the commits.

sushimon (changes requested):
Just make sure your tests actually run with the new entities + see other comments.

src/main/java/sign_up/use_case/SignUpDSGateway.java
reganq: Will do

src/main/java/sign_up/drivers/SignUpDatabaseAccess.java
```diff
@@ -0,0 +1,191 @@
...
+        try {
+            userInfo = readLineByLine(adminFilePath);
+        } catch(Exception e) {
+            System.out.println("Wrong Path");
+            userInfo = new ArrayList<>();
+        }
+        return userInfo.get(1)[0];
+    }
+
+    private ArrayList<String[]> readLineByLine(String filePath) throws Exception {
```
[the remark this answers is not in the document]
reganq: Added a case for CSV validation exception

src/main/java/sign_up/drivers/SignUpDatabaseAccess.java
```diff
@@ -0,0 +1,191 @@
...
+        ArrayList<String[]> list = new ArrayList<>();
+        try (Reader reader = Files.newBufferedReader(Path.of(filePath))) {
```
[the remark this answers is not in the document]
reganq: Removed, it wasn't.

src/test/java/entities/PostTest.java
reganq: I think this one was accidentally committed. I'll remove it

src/main/java/sign_up/drivers/SignUpDataAccess.java
```diff
@@ -0,0 +1,207 @@
...
+
+public class SignUpDataAccess implements DsGateway {
+    private final String userFilePath;
+    private final String adminFilePath;
+    private final String numUsersCreatedFilePath;
+
+    public SignUpDataAccess(String userFilePath, String adminFilePath, String numUsersCreated) {
+        this.userFilePath = userFilePath;
+        this.adminFilePath = adminFilePath;
+        this.numUsersCreatedFilePath = numUsersCreated;
```
sushimon, line 27: Nitpick: why not just name the parameter `numUsersCreatedFilePath` as well?

src/main/java/sign_up/drivers/SignUpDataAccess.java
```diff
@@ -0,0 +1,207 @@
...
+    @Override
+    public int getNumberUsers() {
+        ArrayList<String[]> numUsers;
+        try {
+            numUsers = readLineByLine(numUsersCreatedFilePath);
+        } catch(IOException e) {
+            System.out.println("Wrong Path");
+            String[] defaultEntry = {"0"};
+            numUsers = (ArrayList<String[]>) Arrays.asList(new String[][]{defaultEntry});
+        } catch(CsvValidationException e) {
```
sushimon: Maybe just catch `CsvException` instead. I'm not too familiar with the `CsvException` hierarchy and what subclasses extend it (and can be thrown while parsing a csv file) but it might just be safer to catch the parent exception.

sushimon: Also applies elsewhere where `CsvValidationException` is caught

src/main/java/sign_up/drivers/SignUpDataAccess.java
```diff
@@ -0,0 +1,207 @@
...
+
+    private String convertBoolToString(boolean toConvert) {
+        if (toConvert) {
+            return "1";
+        } else {
+            return "0";
+        }
+    }
+
+    private void write(ArrayList<String[]> toWrite, String filePath) throws Exception {
```
sushimon: Should be more specific as to which exception it throws

src/main/java/sign_up/use_case/DsGateway.java
```diff
@@ -0,0 +1,17 @@
+package sign_up.use_case;
```
sushimon, line 1: Rename to `SignUpDsGateway` so its obvious that its the DsGateway specific to this use case

sushimon (changes requested):

src/main/java/sign_up/drivers/SignUpDataAccess.java
reganq: Changed this

src/main/java/sign_up/drivers/SignUpDataAccess.java
reganq: I think this method has been removed now. I put the uses for it into the specific classes

src/main/java/sign_up/interface_adapters/SignUpPresenter.java
```diff
@@ -0,0 +1,34 @@
...
+    /*
+    *
+    * */
+    @Override
+    public void updateViewModel(SignUpResponseModel responseModel) {
+        if (responseModel.isCreationSuccess()) {
+            viewModel.updateViewModel(new SignUpUserOutputData(responseModel.isCreationSuccess(),
+                    responseModel.getMessage()));
+        } else {
+            viewModel.updateViewModel(new SignUpUserOutputData(responseModel.isCreationSuccess(),
```
izumidayo, line 30: I don't think you need an if/else here because the following code is the same regardless!

BerryRahul (dismissed):
approving to merge

izumidayo (approved):
There's one small part that could be fixed but it shouldn't be causing any errors so lgtm !

PR #31, opened 2022-11-15
Added Make Post UC. 

JohannLeidenfrost:
Added a small change. creationDate is now being initialised as the current date in the Post constructor.


reganq (approved):
lgtm :)

sushimon (dismissed):
lgtm

reganq (approved):
It looks good to me! But one thing, is the MakePostInteractor a work in progress? I guess I can't see the rest of the use case right now.

src/main/java/Database/currentUser.csv
```diff
@@ -0,0 +1 @@
+﻿"currentUserID","isAdmin","email","password","postIDs","favouritesIDs"
```
Thehashhobo, line 1: Do we really need to store currentUser in a csv? if we use it so often, can we just have it in the program after a user is logged in?

BerryRahul: I don't think we do, since CurrentUser will always be set to one specific User or null. If A use case ever needs to pull the currentUser we can just use a getter method. Though currently the getCurrentUser() method in the entity only returns the users email, so that might need to be revised.

JohannLeidenfrost: Yeah makes sense I just added it for now so I could actually use my use case since currentUser has not been implemented!

src/main/java/makePost/drivers/MakePostDatabaseAccess.java
```diff
@@ -0,0 +1,193 @@
...
+import com.opencsv.CSVWriter;
+
+import java.io.*;
+import java.util.*;
+
+public class MakePostDatabaseAccess implements MakePostDataAccessInterface {
+
+    @Override
+    public int getNumberOfPosts(){
+        String filePath = "src/main/java/Database/numPostsCreated.csv";
```
reganq, line 18: I think we should pass the filepaths for different files to the data access as arguments in the constructor. This way, if something needs to change when we are exporting our work, we'll need to change next code

src/main/java/makePost/use_case/MakePostInteractor.java
```diff
@@ -0,0 +1,107 @@
...
+    private final MakePostDataAccessInterface dataAccess;
+    private final MakePostOutputBoundary presenter;
+
+    public MakePostInteractor(MakePostDataAccessInterface dataAccess, MakePostOutputBoundary presenter){
+        this.dataAccess = dataAccess;
+        this.presenter = presenter;
+    }
+
+    /**
+     * Checks whether the deadline is at most six months from the creationDate.
```
reganq, line 23: should we also check that the deadline hasn't passed?

src/main/java/makePost/ui/MakePostView.java
```diff
@@ -0,0 +1,102 @@
...
+        makePostArea.add(tagsLabel);
+        makePostArea.add(tagsScroller);
+        JLabel mainDescLabel = new JLabel("Describe your project: ");
+        makePostArea.add(mainDescLabel);
+        makePostArea.add(enterMainDescription);
+        JLabel deadlineLabel = new JLabel("Enter the date (YYYY-MM-DD) when your post should be deleted (max. six months): ");
+        makePostArea.add(deadlineLabel);
+        makePostArea.add(enterDeadline);
+        makePostArea.add(makePostButton);
+
```
reganq, line 65: I think that because, within makepost view, we create a new panel and a new frame, and then put that panel within the frame and display it, we do not need to extend jpanel with make post view.

reganq (approved):
This looks good to me. I think the way we use the view could use reworking, see my comment in MakePostView. Everything else is good though

src/main/java/makePost/ui/MakePostView.java
JohannLeidenfrost: Fixed this

src/main/java/makePost/use_case/MakePostInteractor.java
JohannLeidenfrost: Done.

src/main/java/makePost/interface_adapters/MakePostController.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public void passToInteractor(Map<String, Object> postBody){
+        int numPostsCreated = interactor.getNumPostsCreated();
+        int currentUserID = interactor.getCurrentUser();
+
+        //raise an error if deadline is not in the right format.
+        List<String> deadlineList = new ArrayList<>(Arrays.asList(((String) postBody.get("deadline")).split("-")));
+        try{
+            if(Integer.parseInt(deadlineList.get(1)) > 12 || Integer.parseInt(deadlineList.get(1)) < 1 || Integer.parseInt(deadlineList.get(2)) > 31
```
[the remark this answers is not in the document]
sushimon: Could refer to the sample user login repo on quercus for help

src/main/java/makePost/use_case/MakePostDataAccessInterface.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+public interface MakePostDataAccessInterface {
+    int getNumberOfPosts();
+    void setNumberOfPosts(int newNumPostsCreated);
+    void savePost(Map<String, String> postAttributes);
+    int getCurrentUser();
+    List<Map<String, Object>> getFavouritedUsers();
```
sushimon, line 13: If we are reading from the database, does the map still have to map from strings to objects? Same thing for `getReplies`.

src/main/java/makePost/use_case/MakePostException.java
```diff
@@ -0,0 +1,14 @@
+package makePost.use_case;
```
sushimon, line 1: It might be better to put this class in a separate folder (e.g. inside `src/main/java/exceptions`).

src/main/java/makePost/use_case/MakePostInputBoundary.java
```diff
@@ -0,0 +1,22 @@
+package makePost.use_case;
+
+import entities.Post;
+import makePost.interface_adapters.MakePostViewModel;
+
+import java.util.List;
+import java.util.Map;
+
+public interface MakePostInputBoundary {
+    void makePost(MakePostRequestModel mprm);
```
sushimon, line 10: IMO, you can rename `mprm` to `requestModel` to make it more understandable.

src/main/java/makePost/use_case/MakePostInputBoundary.java
```diff
@@ -0,0 +1,22 @@
...
+    int getNumPostsCreated();
+    int getCurrentUser();
+    List<Map<String, Object>> getFavouritedUsers();
```
sushimon, line 13: Not sure if this also applies here but I thought I'd ask about the same concern as the comment in `MakePostDataAccessInterface`.

src/main/java/makePost/use_case/MakePostDataAccessInterface.java
```diff
@@ -0,0 +1,22 @@
+package makePost.use_case;
```
sushimon, line 1: Also, could we rename this to `MakePostDsGateway`? We are using the `DsGateway` suffix to denote the data access interface.

src/main/java/makePost/use_case/MakePostInteractor.java
```diff
@@ -0,0 +1,124 @@
...
+     * @return true if the above condition is met.
+     */
+    public boolean checkDeadline(MakePostRequestModel mprm){
```
sushimon, line 23: Could use `requestModel` instead of `mprm` for readability.

src/main/java/makePost/use_case/MakePostInteractor.java
```diff
@@ -0,0 +1,124 @@
...
+        LocalDate deadline = mprm.getDeadline();
+        LocalDate creationDate = LocalDate.now();
+        if((deadline != null) && ((DAYS.between(deadline, creationDate) <= 182.5) && (DAYS.between(deadline, creationDate) >= 0))){
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public void makePost(MakePostRequestModel mprm) {
```
sushimon, line 33: Same comment about renaming this as on line 23

src/main/java/makePost/use_case/MakePostInteractor.java
```diff
@@ -0,0 +1,124 @@
...
+    public int getCurrentPostID() {
+        return this.dataAccess.getCurrentPostID();
+    }
+
+    @Override
+    public String getCreationDate() {
+        return this.dataAccess.getCreationDate();
+    }
+
+    private MakePostResponseModel makePostHelper(MakePostRequestModel mprm) throws MakePostException {
```
sushimon, line 114: Same comment as line 23

src/main/java/makePost/use_case/MakePostInteractor.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    public List<Map<String, Object>> getFavouritedUsers() {
+        return this.dataAccess.getFavouritedUsers();
+    }
+
+    public int getNumPostsCreated(){
+        return this.dataAccess.getNumberOfPosts();
+    }
+
+    public List<Map<String, Object>> getReplies(){
```
sushimon, line 100: Refer to comment in `MakePostDataAccessInterface`

src/test/java/entities/TagTest.java
```diff
@@ -11,15 +11,15 @@
 import static org.junit.Assert.*;
```
sushimon, line 11: Is it possible to uncommit this file? We already deleted it so I don't know what's going to happen once we merge this PR.

sushimon (changes requested):
Notes:
- Was the `tags.csv` file supposed to be for the Tag entity or something else? If it's for the Tag entity, I think this file can be removed.
- Some of your variables aren't in camel case. Make sure to change them and anywhere they are mentioned to camel case.
- There are still some methods in `MakePostDatabaseAccess` that seem incomplete. For example, `getFavouritedUsers`, `getReplies`, etc. Just commenting this for reference.
- Add documentation.
- A lot of mappings from string to objects. Does the value have to be of type `Object` or do we know what type it could be in some classes (e.g. when reading from csv it will be string)? The frequent casting just seems prone to error somewhere.

src/main/java/makePost/drivers/MakePostDatabaseAccess.java
```diff
@@ -0,0 +1,193 @@
...
+        try {
+            // Create an object of filereader
+            // class with CSV file as a parameter.
+            FileReader filereader = new FileReader(file);
+
+            // create csvReader object passing
+            // file reader as a parameter
+            CSVReader csvReader = new CSVReaderBuilder(filereader).withSkipLines(1).build();
+            return Integer.parseInt(csvReader.peek()[0]);
+        } catch (IOException e) {
```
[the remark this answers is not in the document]
JohannLeidenfrost: No, it doesn't ask me to!

sushimon (changes requested):
Notes:
- Might be something that you add later but I think adding an attribute for the file path in `MakePostDatabaseAccess` will help with making testing and future changes easier.
- Still some comments from the last review left unaddressed.
- Also, just make sure the tests run with the new entities.

reganq (dismissed):
lgtm - lets merge

BerryRahul (dismissed):
lgtm

src/main/java/Database/currentUser.csv
JohannLeidenfrost: Okay I changed my code to use the static variable CurrentUser.getCurrentUser() instead!

src/main/java/makePost/drivers/MakePostDatabaseAccess.java
JohannLeidenfrost: Done!

JohannLeidenfrost:
To do:
-add documentation
-move duplicate code and helper code into private methods
-write tests
-make makePostView close upon success

src/main/java/make_post/drivers/MakePostDatabaseAccess.java
```diff
@@ -0,0 +1,139 @@
...
+            CSVReader csvReader = new CSVReader(fileReader);
+            // create CSVWriter object filewriter object as parameter
+            List<String[]> csvBody = csvReader.readAll();
+            csvBody.get(1)[0] = String.valueOf(newNumPostsCreated);
+            for(int i = 0; i < csvBody.size(); i++){
+                if(i >= csvBody.size()){
+                    break;
+                }
+                if(csvBody.get(i).length == 0 && csvBody.get(i)[0].equals("")){
+                    csvBody.remove(i);
```
sushimon, line 58: I think mutating while iterating over a mutable object is dangerous. I think an example list `csvBody = ["hi", ,]` will cause the for loop to not perform its intended task (the resulting list is `csvBody = ["hi", ]` which still contains an empty entry). I think a better choice would be to just create a new list and add the non-empty rows to the new list.

src/main/java/make_post/ui/MakePostView.java
```diff
@@ -0,0 +1,110 @@
...
+    private final JTextField enterDeadline;
+
+    /**
+     * This provides the view for making a Post. It has TextFields for entering all the information necessary for making
+     * a Post.
+     * @param presetTags These are the tags that will be used to filter the Posts.
+     * @param makePostController The controller for the Make Post use case.
+     */
+    public MakePostView(String[] presetTags, MakePostController makePostController){
+        this.TAGS = new String[]{"Sports", "Tech", "Startups"};
```
sushimon, line 32: Should this be assigning `presetTags` to `TAGS` instead? Also, I think `TAGS` should be `private static` instead and just add a getter for it if we need to retrieve it.

sushimon (changes requested):

src/main/java/make_post/ui/MakePostView.java
JohannLeidenfrost: Yes, it should be assigning presetTags (which we instantiate in Main, I think) but since we do not have Main yet, I just used this to check if my code is working. Thanks for flagging this, I completely forgot about it. If I set this to presetTags, then we don't need TAGS anymore.

src/main/java/make_post/drivers/MakePostDatabaseAccess.java
JohannLeidenfrost: Ok I'll update this!

src/main/java/make_post/interface_adapters/MakePostController.java
```diff
@@ -0,0 +1,44 @@
...
+            formatter.withLocale(Locale.getDefault());
+            LocalDate deadline = LocalDate.parse((String) postBody.get("deadline"));
+
+            postBody.put("deadline", deadline);
+            postBody.put("numPostsCreated", numPostsCreated);
+            postBody.put("poster", currentUserID);
+            MakePostRequestModel makePostRequestModel = new MakePostRequestModel(postBody);
+            interactor.makePost(makePostRequestModel);
+        }
+        catch(NumberFormatException | IndexOutOfBoundsException | DateTimeParseException n){
```
reganq, line 39: Are all these exceptions caused by the same thing? may need a diff printout

src/main/java/make_post/use_case/MakePostInteractor.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public void makePost(MakePostRequestModel requestModel) {
+        Post newPost = postFactory.create(requestModel.getPoster(), requestModel.getTitle(), requestModel.getMainDesc(), requestModel.getTags(), requestModel.getCollaborators(),
+                requestModel.getDeadline(), requestModel.getNumPostsCreated());
+        Map<String, String> postAttributes = constructPostAttributes(newPost);
+
+        try{
+            MakePostResponseModel responseModel = this.makePostHelper(requestModel);
+            this.presenter.updateViewModel(responseModel);
+            this.dataAccess.savePost(postAttributes);
```
reganq, line 44: I think this should be done before response model is passed to the presenter

reganq (approved):
Looks good to me!

BerryRahul (approved):
lgtm!

sushimon (approved):

src/main/java/make_post/ui/MakePostView.java
```diff
@@ -0,0 +1,112 @@
...
+        if (evt.getPropertyName().equals("creation failure")) {
+            JFrame errorFrame = new JFrame("Error");
+            switch ((String) evt.getNewValue()) {
+                case ("Deadline more than 6 months away or in the past"):
+                    JOptionPane.showMessageDialog(errorFrame, "Deadline more than 6 months away or in the past",
+                            "Error", JOptionPane.ERROR_MESSAGE);
+                    break;
+                case ("Date is not in the correct format."):
+                    JOptionPane.showMessageDialog(errorFrame, "Date is not in the correct format.",
+                            "Error", JOptionPane.ERROR_MESSAGE);
```
reganq, line 107: you may need to include a "break;" command here

reganq (approved):
I think the break needs to be added in your error popups, but everything else is good!

PR #34, opened 2022-11-19
My Current LogIn Use Case, feedback is appreciated

BerryRahul:
My view is a wip but Im not 100% sure I implemented my presenter/viewmodel correctly

### DIFF
--- a/src/main/java/login/ui/LogInView.java
+++ b/src/main/java/login/ui/LogInView.java
@@ -1,0 +1,110 @@
+package login.ui;
+
+import login.interface_adapters.LogInController;
+import login.interface_adapters.LogInViewModel;
+
+import javax.swing.*;
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+public class LogInView extends JPanel implements ActionListener {
+
+    private final LogInViewModel viewModel;
+
+    private final LogInController controller;
+
+    private final JTextArea enterEmail = new JTextArea("");
+    private final JTextArea enterPassword = new JTextArea("");
+
+
+    public LogInView(LogInController controller, LogInViewModel viewModel){
+        this.controller = controller;
+        this.viewModel = viewModel;
+
+
+        this.setLayout(new GridBagLayout());
+        GridBagConstraints constraints = new GridBagConstraints();
+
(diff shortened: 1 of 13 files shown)


src/main/java/login/ui/LogInView.java
```diff
@@ -0,0 +1,110 @@
...
+        this.add(textLabel, constraints);
+
+        JLabel email = new JLabel("Email");
+        constraints.fill = GridBagConstraints.HORIZONTAL;
+        constraints.weightx = 0.5;
+        constraints.gridx = 3;
+        constraints.gridy = 8;
+        this.add(email, constraints);
+
+
```
reganq, line 43: I think there is a missing password label. At least I don't see it here

src/main/java/login/ui/LogInView.java
```diff
@@ -0,0 +1,110 @@
...
+    public void actionPerformed(ActionEvent e) {
+        String emailInput = enterEmail.getText();
+        String passwordInput = enterPassword.getText();
+
+        enterEmail.setText(" ");
+        enterPassword.setText(" ");
+
+        controller.logInInitializer(emailInput, passwordInput);
+        if (viewModel.getLogInSuccess()){
+            // call main view
```
reganq, line 79: I think the best way to access main view would be to have it as an observable for the view, and tell the observables to update themselves. See Richards or my use case for an example of that.

reganq (commented):
I think it looks good so far. Some things though:
- Your interactions between the view and the viewmodel seem to be ok. However, I think that they may work better if we implement them using the observer design pattern, as I mentioned in a couple review comments. This would mean that the view doesn't directly depend on the viewmodel, even though it wouldn't be a violation of clean architecture. You can see the implementation and the package I used in my use case
- For the testing, you should also try to cover the presenter and the controller, as well as the viewmodel. The view doesn't need to be tested, mainly because we can't do it.  It should be possible to have a single testing file that spans the interface adapters and the use case code
-  The initialization of the viewmodel doesn't seem to need the initialization of the variables, but I know it is a work in progress so not much of a comment there
- More comments and documentation are a good idea, although I think your code is easy to understand as is.
- For the database access, refer to the discord for setting up Opencsv, as you will need the methods for saving, etcetera. I found this turotial helpful: https://www.baeldung.com/opencsv

PR #35, opened 2022-11-19
Comment

Thehashhobo:
updated entities  as discussed in the meeting Thursday


src/main/java/entities/Postable.java
```diff
@@ -11,29 +9,22 @@
...
-     */
-    public void addComment(Comment comment){
-        this.replies.add(comment);
-    }
+    // unique identifier for this postable
+    int id;
 
     /**
-     * Returns the list of replies to a Postable object.
+     * Returns the date of which this Postable was created.
```
sushimon, line 20: Documentation does not match what the method does

src/main/java/entities/Post.java
```diff
@@ -80,11 +80,10 @@ public LocalDate getDeadline(){
     }
```
sushimon, line 81: Could we also create a getter method for title?

sushimon: I added one in my PR for filter post but this will probably get pushed before so its best to add that change here so it gets added earlier

sushimon (commented):
See the comments I left. Also, do the tests in `PostTest` still run and pass after these changes?

src/main/java/entities/Post.java
```diff
@@ -80,11 +80,10 @@ public LocalDate getDeadline(){
...
+     * Returns the list of replies to a Postable object.
      */
-    public List<Postable> getReplies(){
-        return this.replies;
-    }
+    public List<Postable> getReplies(){return this.replies;}
+
     /**
      * adds a reply to the list of replies to this Post.
      * @param reply a reply to this Post.
```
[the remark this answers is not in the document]
Thehashhobo: so I asked Johann who wrote addreply which returns a bool and he said he might need it, add comment does not return anything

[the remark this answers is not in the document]
Thehashhobo: true

src/main/java/entities/Post.java
Thehashhobo: ok i will check it and add it

Thehashhobo:
Btw Richard the test passes after test changes accounting for the modification in the way ids are counted.

src/main/java/entities/User.java
```diff
@@ -29,13 +26,11 @@ public class User {
     *
     * @param admin whether the user will have admin priviledges
     * */
-    public User(boolean admin, int numUsersCreated, String email, String password) {
+    public User(boolean admin, int id, String email, String password) {
         this.favourites = new ArrayList<Post>();
```
izumidayo, line 30: this should be an arraylist of integers (the ids of the posts), please also change the attributes as well thanks

src/main/java/entities/User.java
```diff
@@ -29,13 +26,11 @@ public class User {
...
         this.posts = new ArrayList<Post>();
```
izumidayo, line 31: this should be an arraylist of integers (the ids of the posts), please also change the attributes as well thanks

src/main/java/entities/Post.java
```diff
@@ -80,11 +80,10 @@ public LocalDate getDeadline(){
     }
 
     /**
-     * @return the replies made to this Post.
+     * Returns the list of replies to a Postable object.
      */
-    public List<Postable> getReplies(){
-        return this.replies;
-    }
+    public List<Postable> getReplies(){return this.replies;}
```
izumidayo, line 85: return tyoe should be List<Integers>

src/main/java/entities/Post.java
```diff
@@ -9,33 +9,33 @@
  * description,tags,collaborators, deadline, and users who favourited it.
  */
 public class Post extends Postable{
-    public static int numPostsCreated = 0;
-    /**The id is equal to the number of posts created.*/
-    private int id;
+
     /**This is a list of the users who favourited this post.*/
     private List<User> favouritedUsers;
```
izumidayo, line 14: should be List<Integer> or ArrayList<Integer>

src/main/java/entities/Post.java
```diff
@@ -9,33 +9,33 @@
...
      * @param poster the User who is posting this Post. This is an attribute of the superclass Postable.
      * @param title this Post's title.
      * @param mainDesc this Post's description.
      * @param tags the tags this Post is added to.
      * @param collaborators a description of the type of the collaborators the User who posted this Posts is looking for.
      * @param deadline the date on which the Post is to be deleted from the database.
-     * @param numPostsCreated1 the number of posts created so far.
+     * @param id  unique identifier for this post
      */
     public Post(User poster, String title, String mainDesc, List<String> tags, String collaborators,
```
izumidayo, line 32: this.favouritedUsers needs to be initialized

src/main/java/entities/Comment.java
```diff
@@ -1,36 +1,26 @@
...
     /**
      * Initializes an instance of Comment.
      *
      * @param commenter The user that posted this comment.
      * @param body The text content of the comment.
-     * @param numCommentsCreated The number of comments created in the program so far.
+     * @param id The unique id for this comment
      */
-    public Comment(User commenter, String body, int numCommentsCreated){
+    public Comment(User commenter, String body, int id){
```
izumidayo, line 19: we need a second constructor for Comment that takes parameters:
- User commenter
- String body
- int id
- String creationDate

src/main/java/entities/Postable.java
```diff
@@ -11,29 +9,22 @@
     User user;
```
izumidayo, line 9: this should be changed to int userid

src/main/java/entities/Post.java
izumidayo, line 32: Post also needs a second constructor for when we retrieve data from the csv

izumidayo (changes requested):
made a few comments on changes that need to be implemented based on what we discussed in our team meeting on thursday!

src/main/java/entities/Postable.java
```diff
@@ -11,29 +9,22 @@
...
     // the main text content of the Postable
     String body;
-    // a list of Postable objects created as a reply to this Postable
-    List<Postable> replies;
+
     // the date this Postable was created
-    Date creationDate;
+    String creationDate;
```
reganq, line 14: I think the creationDate attribute should be left as a LocalDate, mainly so we can use the methods to compare datewise if needed.

src/main/java/entities/Post.java
```diff
@@ -9,33 +9,33 @@
...
+                LocalDate deadline, int id){
         super.user = poster;
         super.body = mainDesc;
-        super.replies = new ArrayList<>();
-        numPostsCreated = numPostsCreated1;
-        numPostsCreated++;
-        this.id = numPostsCreated;
+        super.id = id;
+        super.creationDate = new Date().toString();
+        this.replies = new ArrayList<>();
```
reganq, line 38: I think this should contain Integers

src/main/java/entities/Post.java
reganq: Also containing Integers

src/main/java/entities/Post.java
reganq: List of Integers

src/main/java/entities/Comment.java
reganq: @izumidayo I think that the creationDate should be a LocalDate, no? It would allow for date methods if needed.

reganq (commented):
I think we need to standardize the relies, favourite posts, etc into stroing integers, which are the ids of the objects they are referring to.
I also think that we should refactor the attribute declarations and constructors so that the  variables are in the same order (ex we have id, email, password, ... in order in declaration and id, email, pass, ... in constructor for User object) to avoid confusion, as we are going to store them in this way in the database. Also, I think id should come first for each object, because it is important for searching and stuff like that

src/main/java/entities/User.java
Thehashhobo: right thanks for the reminder

src/main/java/entities/User.java
```diff
@@ -29,13 +26,11 @@ public class User {
     *
     * @param admin whether the user will have admin priviledges
     * */
-    public User(boolean admin, int numUsersCreated, String email, String password) {
+    public User(boolean admin, int id, String email, String password) {
```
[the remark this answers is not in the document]
Thehashhobo: done

src/main/java/entities/Post.java
Thehashhobo: I checkED and I have 
this.favouritedUsers = new ArrayList<Integer>(); 
already if that's what you meant

src/main/java/entities/Comment.java
Thehashhobo: I think a second construct would only be necessary if we need to store/edit a comment and put it back into the database which I don't think will happen(comments never get edited since it does not have replies now).

creationDate is a string because when we put posts back into the database, the default initialization of date changes it to the time it was put back in, instead of the actual creation date which we want, since creationDate is from postable, comment is also forced to keep it as a string. If we need to use date methods, I think converting a date formated as a string back into a date would be easy.

src/main/java/entities/Postable.java
Thehashhobo: make sense

src/main/java/entities/Post.java
Thehashhobo: done

Thehashhobo:
we can have a quick talk about whether to have creationDate as string or date

reganq (approved):
Lgtm

src/main/java/entities/Comment.java
```diff
@@ -1,36 +1,26 @@
...
 
     /**
      * Initializes an instance of Comment.
      *
      * @param commenter The user that posted this comment.
      * @param body The text content of the comment.
-     * @param numCommentsCreated The number of comments created in the program so far.
+     * @param id The unique id for this comment
      */
-    public Comment(User commenter, String body, int numCommentsCreated){
```
JohannLeidenfrost, line 26: In all the constructors for all the entities, we decided to only use the IDs of the other entities so instead of User commenter we'll have int commenterID. Could you please change that?

izumidayo (approved):

izumidayo (approved):
looks good for now!

JohannLeidenfrost (approved):
looks good but needs a few changes. But we'll try to get it up now!

PR #36, opened 2022-11-19
Most of Favourite UC

izumidayo:
- most of the Favourite UC
- added classes and interfaces to allow for the reading and writing of csv data for Post and User entities (Post/UserFactory, Post/UserReaderInterface, Post/UserStringFactory, Post/UserWriterInterface)

- missing pieces include
> tests
> implementation of methods that closely interact with the database

========================
feedback on clean architecture and anything you may notice would be very much appreciated!!

### DIFF
--- a/src/main/java/favourite/drivers/DataAccess.java
+++ b/src/main/java/favourite/drivers/DataAccess.java
@@ -1,18 +1,67 @@
 package favourite.drivers;
 
-import favourite.use_case.FavouriteDataAccessInterface;
+import entities.Post;
+import entities.User;
+import favourite.use_case.*;
+
+import java.util.ArrayList;
+
+/**
+ * The DataAccess class in the driver layer implements FavouriteDataAccessInterface.
+ * The main functionality of this class is retrieving and saving data to/from the database.
+ */
 
 public class DataAccess implements FavouriteDataAccessInterface {
-//    @Override
-//    public boolean checkIfFavourited(Post post, User user) {
-//        // check if this user has already favourited this post
-//        // will implement once database stuff is figured out
-//        return false;
-//    }
-//
-//    @Override
-//    public User getUser(){;}
-//
-//    @Override
-//    public Post getPost(){;}
+    // postreader is used to allow class DataAccess to access PostFactory to convert data into
+    // Post entity in getPost method
+    private final PostReaderInterface postreader;
+    // userreader is used to allow class DataAccess to access UserFactory to convert data into
+    // User entity in getUser method
+    private final UserReaderInterface userreader;
+    // postwriter is used to allow class DataAccess to access PostStringFactory to convert Post
+    // entity back into a list of strings to save it in the database in the savePostInfo method
+    private final PostWriterInterface postwriter;
+    // userwriter is used to allow class DataAccess to access UserStringFactory to convert User
+    // entity back into a list of strings to save it in the database in the saveUserInfo method
+    private final UserWriterInterface userwriter;
+
+    /**
+     * Initializes DataAccess
+     *
+     * @param postreader an instance of PostReaderInterface
+     * @param userreader an instance of UserReaderInterface
+     * @param postwriter an instance of PostWriterInterface
+     * @param userwriter an instance of UserWriterInterface
+     */
+    public DataAccess(PostReaderInterface postreader, UserReaderInterface userreader,
+                      PostWriterInterface postwriter, UserWriterInterface userwriter){
+        this.postreader = postreader;
+        this.userreader = userreader;
+        this.postwriter = postwriter;
+        this.userwriter = userwriter;
+    }
+
+    @Override
+    public User getUser(){
+        ArrayList userdata =  new ArrayList();
+        return userreader.readUser(userdata);
+    }
+
+    @Override
+    public Post getPost(int postid){
+        ArrayList postdata = new ArrayList(); // use post id to get the list from the database
+        return postreader.readPost(postdata);
+    }
+
+    @Override
+    public void saveUserInfo(User user) {
+        ArrayList userdata = userwriter.writeuser(user);
+        // save userdata to database, update once csv is set up
+    }
+    @Override
+    public void savePostInfo(Post post, int postid){
+        ArrayList postdata = postwriter.writepost(post);
+        // save postdata to database, update once csv is set up
+    }
+
 }
--- a/src/main/java/favourite/drivers/DataAccess.java
+++ b/src/main/java/favourite/drivers/DataAccess.java
@@ -1,0 +1,18 @@
+package favourite.drivers;
+
+import favourite.use_case.FavouriteDataAccessInterface;
+
+public class DataAccess implements FavouriteDataAccessInterface {
+//    @Override
+//    public boolean checkIfFavourited(Post post, User user) {
+//        // check if this user has already favourited this post
+//        // will implement once database stuff is figured out
+//        return false;
+//    }
+//
+//    @Override
+//    public User getUser(){;}
+//
+//    @Override
+//    public Post getPost(){;}
+}
--- a/src/main/java/favourite/ui/FavouriteView.java
+++ b/src/main/java/favourite/ui/FavouriteView.java
@@ -1,0 +1,37 @@
+package favourite.ui;
+
+import favourite.drivers.DataAccess;
+import favourite.interface_adapters.FavouriteViewModel;
+import favourite.interface_adapters.View;
+
+import javax.swing.*;
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+
+public class FavouriteView implements View {
+    private final FavouriteViewModel viewModel;
+    private JButton favouriteButton = new JButton("Favourite");
+
+    public FavouriteView(){
+        this.viewModel = new FavouriteViewModel(new DataAccess(),this);
+    }
+
+    public void createScreen(Container pane){
+        favouriteButton = new JButton("Favourite");
+        favouriteButton.setBounds(10, 80, 80, 25);
+
+    }
+
+    public void update(){;}
+    public void view(){
+        favouriteButton.addActionListener(new ActionListener() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                viewModel.favourite();
+            }
+        });
+    }
+
+}
--- a/src/main/java/favourite/ui/FavouriteView.java
+++ b/src/main/java/favourite/ui/FavouriteView.java
@@ -3,35 +3,43 @@ package favourite.ui;
 import favourite.drivers.DataAccess;
 import favourite.interface_adapters.FavouriteViewModel;
 import favourite.interface_adapters.View;
+import favourite.use_case.FavouriteResponseModel;
+import favourite.use_case.PostFactory;
 
 import javax.swing.*;
 import java.awt.*;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 
-
-public class FavouriteView implements View {
+/**
+ * The FavouriteView class in the UI layer implements View and ActionListener interfaces.
+ * Whenever a user interacts with the favourite button, this class runs the Favourite use case.
+ */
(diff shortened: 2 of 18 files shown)


src/main/java/favourite/use_case/PostStringFactory.java
```diff
@@ -0,0 +1,16 @@
...
+
+import entities.Post;
+
+import java.util.ArrayList;
+
+/**
+ * PostStringFactory in the use case layer implements PostWriterInterface.
+ * It converts Post objects into a format that the database can save.
+ */
+public class PostStringFactory implements PostWriterInterface {
```
reganq, line 11: A useful way to do this is to use a mapping from the kind of attribute (title, poster, etc) to actual data, and pass that through to the dataaccess and have it interpret this

reganq: Moreover, I don't think we need a 'factory' to do this, as most of the interaction with the data gateway should be done through the interactor. I believe this task could be done as a method in the interactor, which would then pass it to the database, as opposed to having the database call something in the use case layer to do it

src/main/java/favourite/use_case/PostWriterInterface.java
```diff
@@ -0,0 +1,15 @@
+package favourite.use_case;
+
+import entities.Post;
+
+import java.util.ArrayList;
+
+public interface PostWriterInterface {
```
reganq, line 7: I think this is also not needed, as from my point in PostStringFactory

src/main/java/favourite/use_case/UserReaderInterface.java
```diff
@@ -0,0 +1,15 @@
+package favourite.use_case;
+
+import entities.User;
+
+import java.util.ArrayList;
+
+public interface UserReaderInterface {
```
reganq, line 7: These are actually needed though I believe, to allow the data access to 'create' users

src/main/java/favourite/use_case/PostStringFactory.java
reganq: I think the class name should be changed from facotory to avoid confusion, but it does not need the interface IMO as it should just be called from the interactor, and then have its return argument passed to dataaccess

src/main/java/favourite/use_case/UserStringFactory.java
```diff
@@ -0,0 +1,22 @@
...
+
+import entities.User;
+
+import java.util.ArrayList;
+
+/**
+ * UserStringFactory in the use case layer implements UserWriterInterface.
+ * It converts User objects into a format that the database can save.
+ */
+public class UserStringFactory implements UserWriterInterface{
```
reganq, line 11: I don't think this needs the interface, as it should be called by the interactor, similarly to above with the post case

reganq (commented):
I think this is a good start! Some notes, though:
- For the interface adapters, we should change to have a presenter and controller and a viewmodel together, where the controller is accessed by the view and calls the interactor, and the presenter is called by the interactor and updates the viewmodel, which is observed by the view, causing it to change. See mine or richards pull requests for examples
- For your String factories, I think they should be called by the interactor and have their output passed to the data access, as this was the data access will not need to handle user entities as much (barring the reader interfaces)

src/main/java/favourite/use_case/FavouriteDataAccessInterface.java
```diff
@@ -0,0 +1,35 @@
+package favourite.use_case;
+
+import entities.Post;
+import entities.User;
+
+public interface FavouriteDataAccessInterface {
```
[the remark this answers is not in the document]
izumidayo: got it! thanks for pointing this out

src/main/java/database/posts.csv
```diff
@@ -0,0 +1 @@
+postID,"posterID","title","mainDescription","tags","collaborators","deadline","creationDate", favouritedUsersIDs,"repliesIDs";1,2,My new and shiny project,This is the description of my project. Who wants to join me?,tag1 tag2,,2022 12 25,2022 11 11,1 4,;2,4,I want to start a Quant company,This is the description of my project. Who wants to join me?,"",,2022 12 25,2022 11 11,1,;3,3,Join my dance club!,This is the description of my project. Who wants to join me?,tag3,,2022 12 25,2022 11 11,2,;4,2,UofT Merch Company,This is the description of my project. Who wants to join me?,tag1 tag3,,2022 12 25,2022 11 11,4,;5,1,UofT Grey's Anatomy fan community,This is the description of my project. Who wants to join me?,tag1,,2022 12 25,2022 11 11,,
```
JohannLeidenfrost, line 1: Can you remove this posts.csv from your pull requests or at least remove the test data that you uploaded here? This is so we have a file with the column names but no actual data on main!

src/main/java/database/users.csv
```diff
@@ -0,0 +1 @@
+ userID,isAdmin,email,password,listPosts,listFavourites;1,TRUE,email1@mail.utoronto.ca,12345,5,1 2;2,FALSE,email2@mail.utoronto.ca,PasSwOrd,4 1,3;3,TRUE,email3@mail.utoronto.ca,iloveapples,3,;4,FALSE,email4@mail.utoronto.ca,482js8,2,1 4
```
JohannLeidenfrost, line 1: Please remove the test data from here!

src/main/java/entities/Comment.java
```diff
@@ -1,55 +0,0 @@
-package entities;
-
```
JohannLeidenfrost, line 2: Why did you delete the comment entity?

src/main/java/favourite/drivers/DataAccess.java
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * The DataAccess class in the driver layer implements FavouriteDataAccessInterface.
+ * The main functionality of this class is retrieving and saving data to/from the database.
+ */
+
+public class DataAccess implements FavouriteDSGateway {
+    // postreader is used to allow class DataAccess to access PostFactory to convert data into
+    // Post entity in getPost method
+    private final PostReaderInterface postreader;
```
JohannLeidenfrost, line 28: Can you use camel case in postreader?

src/main/java/favourite/drivers/DataAccess.java
```diff
@@ -0,0 +1,192 @@
...
+    // userreader is used to allow class DataAccess to access UserFactory to convert data into
+    // User entity in getUser method
+    private final UserReaderInterface userreader;
+    // partialpath is a String that contains the partial directory that leads to all csv files in the program
+    // it is isolated from the file name so that in case the file is moved, the code is still compatible
+    private final String partialpath;
```
JohannLeidenfrost, line 34: camel case for all these variables!

src/main/java/favourite/ui/FavouriteView.java
```diff
@@ -0,0 +1,62 @@
...
+    private final FavouriteController controller;
+    // the id of the post being favourited/unfavourited
+    private final int postid;
+
+    /**
+     * Initializes FavouriteView
+     *
+     * @param postid id of the post being favourited/unfavourited
+     */
+    public FavouriteView(int postid, FavouriteController controller){
```
reganq, line 26: The int being passed as a param in constructor seems to be valid. This requires that the MainView knows about all the controllers (favourite, makepost, comment, etc), however

src/main/java/favourite/use_case/FavouriteInteractor.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Favourites a post if the user has not already, unfavourites a post if the user has favourited it already.
+     *
+     * @param requestModel tells the FavouriteInteractor which post is being favourited/unfavourited
+     */
+    @Override
+    public void favouritepost(FavouriteRequestModel requestModel) {
+        int userid = CurrentUser.getCurrentUser();
```
reganq, line 35: Will need to change this to work with changes to currentuser

reganq (dismissed):
I think you should change the constructor for the controller, but besides that its good!

src/main/java/favourite/interface_adapters/FavouriteController.java
```diff
@@ -0,0 +1,32 @@
...
+public class FavouriteController{
+    private final FavouriteInputBoundary interactor;
+    private final int postID;
+
+    /**
+     * Initializes FavouriteController
+     * @param interactor the interactor, passing as a FavouriteInputBoundary
+     * @param postID the id of the post being favourtied/unfavourited
+     */
+    public FavouriteController (FavouriteInputBoundary interactor, int postID){
```
JohannLeidenfrost, line 15: Does the controller need to have postID in the constructor?

src/main/java/favourite/use_case/FavouriteDSGateway.java
```diff
@@ -0,0 +1,36 @@
+package favourite.use_case;
+
+import entities.Post;
+import entities.User;
+
+public interface FavouriteDSGateway {
```
JohannLeidenfrost, line 6: minor detail: I think we use the convention DsGateway instead of DSGateway

src/main/java/favourite/use_case/PostFactory.java
```diff
@@ -0,0 +1,63 @@
...
+        String mainDesc = postdata[3];
+        String[] tagarray = postdata[4].split(" ");
+        List<String> tags = new ArrayList<>();
+        for (String tag : tagarray) {
+            tags.add(tag);
+        }
+        String collaborators = postdata[5];
+        // creating LocalDate object for deadline
+        String[] deadlineElements = postdata[6].split(" ");
+        int year1 = Integer.parseInt(deadlineElements[0]);
```
JohannLeidenfrost, line 34: To make your life easier here, you can use LocalDate.parse(String date) instead! The date has to be in the format yyyy-MM-dd though, which should be the case in the database! It will automatically throw an error if the date does not exist, e.g. February 31st

src/test/java/favourite/users.csv
```diff
@@ -0,0 +1,5 @@
+userID,isAdmin,email,password,listPosts,listFavourites
+1,TRUE,email1@mail.utoronto.ca,12345,5,1 2
```
JohannLeidenfrost, line 2: remove test data from here please!

JohannLeidenfrost (changes requested):
Looks good except a couple of quick fixes:
-remove test data from all csv files
-updates some variable names to make them camel case if you have time 
-look at the FavouriteController and check whether the postID is necessary in its constructor

src/main/java/favourite/interface_adapters/FavouriteController.java
izumidayo: Good point, fixed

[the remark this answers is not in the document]
izumidayo: thanks for the feedback, would love to discuss in tutorial today

src/main/java/database/posts.csv
izumidayo: fixed!

src/main/java/database/users.csv
izumidayo: fixed!

src/main/java/entities/Comment.java
izumidayo: I don't recall doing that... it would be great if we could discuss how to resolve this issue in tutorial (whether I should add it back into my branch or let it merge as is)

src/main/java/favourite/drivers/DataAccess.java
izumidayo: fixed!

src/main/java/favourite/drivers/DataAccess.java
izumidayo: fixed"

src/test/java/favourite/users.csv
izumidayo: done!

src/main/java/favourite/use_case/PostFactory.java
izumidayo: this is great, thank you

PR #39, opened 2022-11-20
Current Implementation Of Login and LogOut UC 

BerryRahul:
Reworked my log in interface adapters and view to follow the 
observer design pattern. and added my logout use case for review.


src/main/java/LogOut/ui/LogOutView.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public void propertyChange(PropertyChangeEvent evt) {
+        if (evt.getPropertyName().equals("Log Out")){
+            String logOut = "Log Out";
+            LogOutController logOutController = new LogOutController(logOut);
+            controller.LogOutInitializer(logOutController);
+
+
+            CardLayout cardLayout = (CardLayout) this.getLayout();
```
BerryRahul, line 26: Im not sure if this is the correct way to get the main page view (not sure what it's called) and then show the main login/signup view

src/main/java/LogOut/ui/LogOutView.java
```diff
@@ -0,0 +1,30 @@
+package LogOut.ui;
+
+import LogOut.interface_adapters.LogOutController;
+
+import javax.swing.*;
+import java.awt.*;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+
+public class LogOutView extends JPanel implements PropertyChangeListener {
```
izumidayo, line 10: I have a similar view on my Favourite UC (FavouriteView) where I add a JButton on my view which is a subclass of JPanel like yours. I think if you add a JButton with an actionlistener the way I did, it will be easier to integrate your view with the main view! (Assuming that my method actually works...)

src/main/java/LogOut/ui/LogOutView.java
izumidayo: hmm I'm not quite sure either...

izumidayo (changes requested):
Looks pretty good to me! There's one easy but important fix I think you need to make for your Log In UC so please check that one out. (Implementing LogInOutputBoundary)

BerryRahul:
The LogOut UC should be all fine, just trying to figuring out if my view implementation for LogOut is correct.

src/main/java/logIn/drivers/LogInDatabaseAccess.java
```diff
@@ -0,0 +1,38 @@
...
+import entities.User;
+
+import logIn.use_case.LogInDsGateway;
+
+public class LogInDatabaseAccess implements LogInDsGateway {
+
+    public LogInDatabaseAccess(){
+    }
+
+
```
reganq, line 12: It is possible to set up your db access with the projected csv files. If you look at my sign up use case, you can see that I have the csv for user projected to have its columns like (favourites, posts, isadmin, password, email, id). You should try setting up the csv library and implementing your code in a way similar to what I have in sign up. Let me know if you need help setting up

src/main/java/logIn/exceptions/UserException.java
```diff
@@ -0,0 +1,14 @@
+package logIn.exceptions;
+
```
reganq, line 2: I'm not sure if exceptions should have its own package like this, as it implies it is a layer by itself.

src/main/java/logIn/interface_adapters/LogInPresenter.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public LogInPresenter(LogInViewModel viewModel){
+        this.viewModel = viewModel;
+
+    }
+
+
+    @Override
+    public void present(LogInResponseModel responseModel) {
+        if (responseModel.getLogInSuccess()){
```
reganq, line 18: I realize that I have similar code in my file, but I think this is redundant. could just pass it without if/else

src/main/java/LogOut/ui/LogOutView.java
reganq: The switch here will be done in the main frame i think, as we are switching the card with the main view to the card with the landing view. I did not cover the converse of this in my use case, as we will need to implement the main view, which is just a frame.

reganq (approved):
Everything here looks good, barring the db implementation which I think you still need to add and the view for the log out, which should be refactored to edit the viewmodel of the main frame view, which hasn't been implemented yet. However, the inner code is good, so I'm approving on the merit of that. It should run with or without the db and view, which we can refactor later.
Also see my comment on the exceptions package

sushimon (changes requested):
Notes:
- Can you rename your use case directory to `log_out` instead of `LogOut` and `log_in` instead of `logIn`?

src/main/java/logIn/drivers/LogInDatabaseAccess.java
```diff
@@ -0,0 +1,126 @@
...
+            ArrayList<String> emails = this.getData(4);
+            ArrayList<String> admins = this.getData(2);
+
+            int emailIndex = emails.indexOf(email);
+            String adminValueString = admins.get(emailIndex);
+
+            boolean isAdmin;
+            isAdmin = adminValueString.equals("true");
+
+            return new User(isAdmin, 0, email, pass);
```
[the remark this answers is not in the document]
BerryRahul: is it possible to pull the user object itself from the csv? 
otherwise nice catch I agree I'll change this

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -0,0 +1,126 @@
...
+
+    }
+
+
+    /**
+     * This method will return a List of all the values in a csv file
+     * @param indexInfo the attribute of users in interest
+     * @return a List of a specific attribute of all users in a csv file
+     */
+    public ArrayList<String> getData(int indexInfo){
```
reganq, line 96: I think this can be a private method

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -0,0 +1,126 @@
...
+
+        return data;
+    }
+
+    private ArrayList<String[]> readLine(String path) throws Exception {
+        ArrayList<String[]> list = new ArrayList<>();
+        try (Reader reader = Files.newBufferedReader(Path.of(path))) {
+            try (CSVReader csvReader = new CSVReader(reader)) {
+                String[] line;
+                while ((line = csvReader.readNext()) != null) {
```
reganq, line 119: We may want to skip the header in the csv file (i think they have one). If not, disregard this.

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -0,0 +1,126 @@
...
+     * @return boolean of if the email exists in the database or not.
+     */
+    @Override
+    public boolean checkUserEmailExists(String email){
+        ArrayList<String> emails = this.getData(4);
+        for (int i = 0; i < emails.size(); i++){
+            if (emails.get(i).equals(email)){
+                return true;
+            } else {
+                i++;
```
reganq, line 33: I don't think you need to increment i here, it will be done automatically by the for loop

src/main/java/log_out/ui/LogOutView.java
```diff
@@ -19,12 +19,13 @@ public LogOutView(LogOutController controller){
     public void propertyChange(PropertyChangeEvent evt) {
```
reganq, line 19: Not for this line particular, but the log out view should create a JButton with an action listener to detect being clicked and calling the log out controller

reganq (approved):
Mostly looking good! I noticed a couple things in the db that need to be fixed, and I think the log out view needs a button added, but everything else looks good!

src/main/java/logIn/drivers/LogInDatabaseAccess.java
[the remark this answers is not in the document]
BerryRahul: changed this

src/main/java/log_in/drivers/LogInDatabaseAccess.java
BerryRahul: right, thanks

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -26,12 +28,8 @@ public LogInDatabaseAccess(String userPath){
...
     public boolean checkUserEmailExists(String email){
         ArrayList<String> emails = this.getData(4);
-        for (int i = 0; i < emails.size(); i++){
-            if (emails.get(i).equals(email)){
-                return true;
-            } else {
-                i++;
-            }
+        for (String s: emails){
+            return s.equals(email);
```
reganq, line 32: This only checks the first string

reganq (approved):
With the for loop changes and the actionlistener stuff, I think this is good!

src/main/java/log_in/interface_adapters/LogInControllerData.java
```diff
@@ -0,0 +1,20 @@
+package log_in.interface_adapters;
+
+// inputted data to be processed
+public class LogInControllerData {
```
izumidayo, line 4: I think a few of us have data structures like this in our interface adapters so let's try to decide on a naming convention when we meet next time!

src/main/java/log_out/ui/LogOutView.java
```diff
@@ -0,0 +1,48 @@
...
+    public LogOutView(LogOutController controller){
+        this.controller = controller;
+
+        logOutButton.addActionListener(this);
+    }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent evt) {
+        if (evt.getPropertyName().equals("Log Out")){
+            String logOut = "log out";
```
izumidayo, line 28: not sure what this code is supposed to lead to, but from what I imagine a successful log out would look like, the main view would close and the log in page should be brought up again maybe?

izumidayo (approved):
I left 2 comments but they more so ideas/suggestions we can consider later rather than necessary fixes for errors. Everything else looks good to me!

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Method will call all emails in the database and check if the email given in the parameter
+     * is in the database.
+     * @param email email to be checked in the database.
+     * @return boolean of if the email exists in the database or not.
+     */
+    @Override
+    public boolean checkUserEmailExists(String email){
+        ArrayList<String> emails = this.getData(4);
+        for (int i = 0 ; i < emails.size(); i++ ){
```
JohannLeidenfrost, line 31: You can skip the first line here since the first line are just headers

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -0,0 +1,128 @@
...
+        } else {
+            return null;
+        }
+
+        return userInfo;
+    }
+
+    //testing method see LogInTest for usage
+    @Override
+    public void addUser(User user){
```
JohannLeidenfrost, line 87: Is it necessary to have this method here as well as in the LogInTest?

JohannLeidenfrost (dismissed):
Looks good to me!

sushimon (approved):

reganq (approved):
lgtm

JohannLeidenfrost (approved):
Looks good!

PR #42, opened 2022-11-20
37 entities all entities

JohannLeidenfrost:
I updated the entities so that all the constructors for the entities now only contain the IDs of other entities instead of the entities themselves. I also added a second constructor to both comment and user for when we are retrieving comments and users from the DB.


reganq (approved):
lgtm

sushimon (approved):

izumidayo (approved):

PR #43, opened 2022-11-20
updated with suggestions from pull request

Thehashhobo:
updated test cases and made attributes of post and user id based. Did not change creationdate but I am down to discuss it, as I am also on the edge


sushimon:
You can delete this PR, we resolved the coupling issues in #42 which uses ids to identify users, posts, etc.

Thehashhobo:
I see

PR #45, opened 2022-11-20
Create the view post use case

sushimon:
Still a WIP.

TODO:
- Add tests
- Implement UI stuff
- Configure database stuff


sushimon:
It committed the filter post use case files for some reason

PR #46, opened 2022-11-20
Create the view post use case.

sushimon:
Still a WIP

TODO:
- Add tests
- Implement UI and database stuff

### DIFF
--- a/src/main/java/view_post/use_case/ViewPostInteractor.java
+++ b/src/main/java/view_post/use_case/ViewPostInteractor.java
@@ -1,0 +1,34 @@
+package view_post.use_case;
+
+public class ViewPostInteractor implements ViewPostInputBoundary {
+    private final ViewPostDsGateway dsGateway;
+    private final ViewPostOutputBoundary presenter;
+
+    /**
+     * Initialize the interactor for this use case.
+     *
+     * @param dsGateway The gateway to access the data repository.
+     * @param presenter The presenter of this use case.
+     */
+    public ViewPostInteractor(ViewPostDsGateway dsGateway, ViewPostOutputBoundary presenter) {
+        this.dsGateway = dsGateway;
+        this.presenter = presenter;
+    }
+
+    /**
+     * Display the selected post.
+     * @param requestModel The input data from the user.
+     */
+    @Override
+    public void displayPost(ViewPostRequestModel requestModel) {
+        int postID = requestModel.getPostID();
+        String[] postInfo = dsGateway.getPostInfo(postID);
+        String delimiter = "##";
+
+        ViewPostResponseModel outputData = new ViewPostResponseModel(
+                postInfo[0], postInfo[1], postInfo[2].split(delimiter), postInfo[3].split(delimiter), postInfo[4]
+                );
+
+        presenter.updateActivePost(outputData);
+    }
+}
--- a/src/main/java/view_post/use_case/ViewPostResponseModel.java
+++ b/src/main/java/view_post/use_case/ViewPostResponseModel.java
@@ -1,0 +1,64 @@
+package view_post.use_case;
+
+/**
+ * A data class for storing the output data.
+ */
+public class ViewPostResponseModel {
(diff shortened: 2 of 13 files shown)


src/main/java/view_post/use_case/ViewPostInteractor.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Display the selected post.
+     * @param requestModel The input data from the user.
+     */
+    @Override
+    public void displayPost(ViewPostRequestModel requestModel) {
+        int postID = requestModel.getPostID();
+        String[] postInfo = dsGateway.getPostInfo(postID);
+        String delimiter = "##";
```
reganq, line 26: I could be wrong, but I feel that the delimiter stuff should be in the view post gateway. Mainly just because the delimiter stuff is concerned with the specifics of the database, and we want to hide that from the interactor.

src/main/java/view_post/use_case/ViewPostResponseModel.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * A data class for storing the output data.
+ */
+public class ViewPostResponseModel {
+    private final String posterEmail;
+    private final String postBody;
+    private final String[] postTags;
+    private final String[] postReplies;
+    private final String deadline;
+
```
izumidayo, line 12: Could we add creationDate (String), collaborators (String), and postID (int)?

src/main/java/view_post/use_case/ViewPostResponseModel.java
```diff
@@ -0,0 +1,64 @@
+package view_post.use_case;
+
+/**
+ * A data class for storing the output data.
+ */
+public class ViewPostResponseModel {
+    private final String posterEmail;
+    private final String postBody;
+    private final String[] postTags;
```
izumidayo, line 9: can we have postTags be a String instead?

reganq (approved):
This looks good to push to me. My comment about the delimiters is a thing that can be fixed later.

izumidayo (approved):
Thanks for working on this Richard! I'm going to approve this for now, because I can implement the changes I commented about if this gets merged soon. If it does not get merged in a few hours, it would be great if you could commit the changes I commented about in the meantime.

src/main/java/view_post/use_case/ViewPostResponseModel.java
```diff
@@ -0,0 +1,64 @@
...
+    private final String[] postReplies;
```
izumidayo, line 10: Can we also change this to an array of integers?

src/main/java/view_post/use_case/ViewPostInteractor.java
sushimon: I was thinking that the gateway's job is to just parse the database and get the post info and return it as an array of strings. I was planning on making the DS gateway parse it and add the delimiter so I can turn it into an array of strings in this method.
It's to avoid having the `getPostInfo` method return an array of objects and typecasting instead of just an array of strings.

src/main/java/view_post/use_case/ViewPostResponseModel.java
sushimon: Will add creationDate and collaborators but do we really need postID? This is information that is being displayed to the user and I don't think that piece of info is important to them so we can leave it out. Or are we keeping the postID so we know which post to delete when deleting posts?

src/main/java/view_post/use_case/ViewPostResponseModel.java
sushimon: For representing the ids of the comments right? Will do

src/main/java/view_post/use_case/ViewPostResponseModel.java
sushimon: Will do

JohannLeidenfrost (approved):
Approving so we can get our main view working asap!

PR #47, opened 2022-11-20
Feature 3 deadline timer

reganq:
My pull request for the deadline timer. It is a WIP but nearly done. I am waiting for the delete post use case, as it is the same functionality and it will save time to use it.
Also, I am waiting for the main view viewmodel to be posted, as that is the viewmodel that my presenter needs to manipulate. 
Other than these things I think the use case is ok


src/main/java/deadline_timer/drivers/TimerDataAccess.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+        for (String[] postInfo: allPosts) {
+            HashMap<String, Object> hashMap = convertToHash(postInfo);
+            toReturn.put((Integer) hashMap.get("postID"), (Post) postReader.readPost(hashMap));
+        }
+
+        return toReturn;
+    }
+
+    private ArrayList<String[]> readLineByLine(String filePath) throws Exception{
```
sushimon, line 42: Better practice to throw a more specific exception than a very general one

sushimon (approved):
Just going to approve this so we can get something working soon, but these comments can be for future reference.

Comments/concerns:
- Minor naming convention change but can you change `TimerDSGateway` to `TimerDsGateway`. We're using the `DsGateway` suffix for the data access interface
- IMO its better to throw a more specific exception than a very general one like the entire `Exception` class
- I still don't get why we have `PostReader` and `PostFactory` extending it. If you're worried about the data access skipping a layer and knowing about the entities, why not just format the data in the data access class, then return it so then the interactor can then work with it? Why not inject it into the interactor instead? It seems that it just moves the coupling elsewhere.

reganq:
Going to cut this feature

PR #49, opened 2022-11-21
Added UI and Interface Adapter Elements to View Post

izumidayo:
Still a WIP but added UI and Interface Adapter classes to Richard's earlier PR for View Post. When reviewing the code, please refer to the main code because Richard's earlier work has already been merged and is not displayed in the "files changed" section of thie PR. Sorry about the inconvenience!! :|

Would Appreciate FB on:
- the construction of the main "landing page" view that is made up of 3 subpanels

===========
Parts that still need work
- tests
- scrollable view for list of posts
- ViewPostGateway


reganq (approved):
This looks good! I would recommend you use the observer pattern going forward. Also, I think we need a controller and a presenter, and then having them interact with the viewmodel

sushimon (changes requested):
Could you implement MVVM using the observer pattern found in the [design pattern repo](https://github.com/CSC207-UofT/design-pattern-samples/tree/main/src/observer/withObserverJava9AndLater) instead?

PR #50, opened 2022-11-23
2 feature 4 delete post

lisan-y:
delete post use case.
still working on database and tests.


src/main/java/delete_post/use_case/DeletePostInteractor.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public DeletePostInteractor(DeletePostPresenter presenter, DeletePostDsGateway dataAccess){
+        this.presenter = presenter;
+        this.dataAccess = dataAccess;
+    }
+
+    @Override
+    public void delete(DeletePostRequestModel requestModel){
+        if (requestModel.getIsTimer() || requestModel.getIsAdmin() ||
+                requestModel.getUser() == (requestModel.getPostUser())){
```
reganq, line 19: In this line, the user that created the post should be retreived from the database first

src/main/java/delete_post/use_case/DeletePostInteractor.java
```diff
@@ -0,0 +1,37 @@
...
+            DeletePostResponseModel responseModel = new DeletePostResponseModel(requestModel.getPost());
+
+            for (User favouriteId: responseModel.getFavourites()){
+                this.dataAccess.removeFavourite(responseModel.getId(), favouriteId);
+            }
+
+            for (String tag: responseModel.getTags()){
+                this.dataAccess.removeTag(responseModel.getId(), tag);
+            }
+
```
reganq, line 30: These for loops could be made into helpers - not a pressing matter tho

src/main/java/delete_post/interface_adapters/DeletePostViewModel.java
```diff
@@ -0,0 +1,4 @@
+package delete_post.interface_adapters;
+
```
reganq, line 2: I don't think a viewmodel is necessary for this use case, it seems like it will mostly be interacting w the view post viewmodel

src/main/java/delete_post/use_case/DeletePostResponseModel.java
```diff
@@ -0,0 +1,32 @@
...
+import entities.User;
+
+import java.util.List;
+
+public class DeletePostResponseModel {
+
+    private Post post;
+
+    public DeletePostResponseModel(Post post){
+        this.post = post;
```
reganq, line 14: The response model also likely doesn't need the actual Post object - just the id.

reganq (approved):
*** Please don't commit anything - lets get this approved first***
Most of this looks good, the only think I take issue with is the fact that you still have the request and response models dealing with Posts. I think we can get rid of that feature.
Also, I think you could make specific cases with your errors, so it has a more customized popup when you give an error message.
Approving because I need this stuff for timer use ccase.

src/main/java/delete_post/use_case/DeletePostInteractor.java
```diff
@@ -0,0 +1,37 @@
...
+    private final DeletePostDsGateway dataAccess;
+
+    public DeletePostInteractor(DeletePostPresenter presenter, DeletePostDsGateway dataAccess){
+        this.presenter = presenter;
+        this.dataAccess = dataAccess;
+    }
+
+    @Override
+    public void delete(DeletePostRequestModel requestModel){
+        if (requestModel.getIsTimer() || requestModel.getIsAdmin() ||
```
reganq, line 18: Important - the delete post interactor cannot call the presenter if it has been called by the timer. That way, we can have the two cases completely separate while using the same functionality

BerryRahul (dismissed):
Approving to merge

JohannLeidenfrost (approved):

reganq (approved):
approving - had to resolve merge conflict

BerryRahul (approved):

src/main/java/delete_post/use_case/DeletePostInteractor.java
```diff
@@ -0,0 +1,37 @@
...
+                requestModel.getUser() == (requestModel.getPostUser())){
+            // delete post
+            DeletePostResponseModel responseModel = new DeletePostResponseModel(requestModel.getPost());
+
+            for (User favouriteId: responseModel.getFavourites()){
+                this.dataAccess.removeFavourite(responseModel.getId(), favouriteId);
```
izumidayo, line 24: The interactor shouldn't have to use the response model's getters if it has the ability to construct the response model in the first place. For example, responseModel.getID() gives the same result as requestModel.getPost().getID()

src/main/java/delete_post/interface_adapters/DeletePostPresenter.java
```diff
@@ -0,0 +1,21 @@
...
+public class DeletePostPresenter implements DeletePostOutputBoundary{
+
+    private final DeletePostViewModel viewModel;
+
+    public DeletePostPresenter(DeletePostViewModel viewModel){
+        this.viewModel = viewModel;
+    }
+    @Override
+    public void prepareFailView(String error){
+        //error popup, no change to posts view or database;
```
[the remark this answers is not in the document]
izumidayo: I also second this!

izumidayo (approved):
I left a few comments regarding Clean Architecture and design choices. However, I will approve this PR so that someone can help make changes to the code after merging.

PR #55, opened 2022-11-27
Sign Up UC

reganq:
Fixed the merge conflict (I think) and looking to push to main now.


izumidayo (approved):
looks good to me!

src/main/java/sign_up/drivers/SignUpDatabaseAccess.java
```diff
@@ -0,0 +1,190 @@
...
+                    csvBody.remove(i);
+                    break;
+                }
+            }
+            FileWriter outputFile = new FileWriter(file);
+            CSVWriter writer = new CSVWriter(outputFile);
+            writer.writeAll(csvBody);
+            writer.flush();
+            // closing writer connection
+            writer.close();
```
JohannLeidenfrost, line 70: You could close the reader here as well.

src/main/java/sign_up/drivers/SignUpDatabaseAccess.java
```diff
@@ -0,0 +1,190 @@
...
+            if (list.indexOf(integer) == 0) {
+                toReturn = integer.toString().concat(toReturn);
+            } else {
+                toReturn = integer.toString().concat(" ").concat(toReturn);
+            }
+        }
+        return toReturn;
+    }
+
+    private String convertDateToString(LocalDate toConvert) {
```
JohannLeidenfrost, line 179: You can use LocalDate.parse(String date) instead! The date has to be in the format yyyy-MM-dd though. It will automatically throw an error if the date does not exist, e.g. February 31st

src/main/java/sign_up/use_case/SignUpInteractor.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        if (this.checkEmailExists(requestModel.getEmail())) {
+            throw new MakeUserException("email exists");
+        }
+        if (requestModel.isCheckAdmin() && !this.checkAdminPassword(requestModel.getAdminPassword())) {
+            throw new MakeUserException("admin password");
+        }
+
+        int numUsers = this.dataAccess.getNumberUsers();
+        User newUser = this.createUser(requestModel.isCheckAdmin(), numUsers, requestModel.getEmail(), requestModel.getPassword());
```
JohannLeidenfrost, line 64: I have the same issue but maybe we should reduce the number of parameters in the constructor for user and add all the attributed in a map instead.

src/main/java/sign_up/use_case/SignUpRequestModel.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /*
+    * Initializes a SignUpResponseModel
+    *
+    * @param email String representation of the inputted email
+    * @param password String representation of the inputted password
+    * @param checkAdmin boolean rep of whether to check adminPassword
+    * @param adminPassword String rep of admin password to check
+    */
+    public SignUpRequestModel(String email, String password, boolean checkAdmin, String adminPassword) {
```
JohannLeidenfrost, line 24: same thing about the constructor here!

JohannLeidenfrost (approved):
Looks great! This is definitely ready to merge if there are no errors!

BerryRahul (approved):
looks good to merge

PR #61, opened 2022-11-28
2 feature 4 delete post

lisan-y:
fixed comments from previous pull request.
database done.
still fixing up view and presenter
need to do tests.


build.gradle
```diff
@@ -7,7 +7,6 @@ repositories {
 }
 
 dependencies {
-    implementation 'com.opencsv:opencsv:5.7.1'
```
JohannLeidenfrost, line 10: I don't think you should delete this. GitHub auto grading is showing a bunch of errors now because it cannot find the OpenCSV libraries, etc.

src/main/java/delete_post/drivers/DeletePostDataAccess.java
```diff
@@ -2,24 +2,120 @@
...
+            }
+        }
+        fileWriter(userPath, userData);
     }
     public void deletePost(int postId){
 
+        List<String[]> postData = readFile(postPath);
+
+        int rowNum = 0;
+        for (String[] row: postData){
```
JohannLeidenfrost, line 72: Modifying the length of a list while iterating over it is dangerous. You could instead make a new list and add all the posts except the one you want to remove.

src/main/java/delete_post/interface_adapters/DeletePostPresenter.java
```diff
@@ -2,20 +2,22 @@
...
     private final DeletePostViewModel viewModel;
+    private final Jframe postView;
 
     public DeletePostPresenter(DeletePostViewModel viewModel){
         this.viewModel = viewModel;
     }
     @Override
-    public void prepareFailView(String error){
-        //error popup, no change to posts view or database;
+    public void prepareFailView(){
```
JohannLeidenfrost, line 16: I am not entirely sure but I think here, too, you should update the viewModel (like in prepareSuccessView), which then fires a property change and then the view opens the error view, not the presenter itself. That would be more in keeping with the observer pattern we are using.

src/main/java/delete_post/use_case/DeletePostInteractor.java
```diff
@@ -1,37 +1,37 @@
...
     @Override
     public void delete(DeletePostRequestModel requestModel){
-        if (requestModel.getIsTimer() || requestModel.getIsAdmin() ||
-                requestModel.getUser() == (requestModel.getPostUser())){
+        if (requestModel.getIsTimer()){
+            DeletePostResponseModel responseModel = new DeletePostResponseModel(requestModel.getPostId());
+
+            this.dataAccess.removeFavourites(requestModel.getPostId());
+            this.dataAccess.deletePost(requestModel.getPostId());
+            this.outputBoundary.prepareTimerView(responseModel);
```
JohannLeidenfrost, line 22: Minor detail: to reduce the amount of duplicate code, you could combine both if statements into one so that you have if(requestModel.getIsTimer() || requestModel.getIsAdmin() ||
                 requestModel.getUserId() == this.dataAccess.getPostUser(requestModel.getPostId())){
...
if(requestModel.getIsTimer()){
this.outputBoundary.prepareTimerView(responseModel);
}
else{
this.outputBoundary.prepareSuccessView(responseModel);
}}
Since the only difference between if it is the timer who calls delete post or if it is an admin/user is the view that is called in the end, right?

JohannLeidenfrost (changes requested):
Looks good but please look at my comments, especially the one in the build.gradle file and in DeletePostDataAccess.

src/main/java/delete_post/interface_adapters/DeletePostPresenter.java
```diff
@@ -2,20 +2,22 @@
...
+        DeleteErrorView(postView);
```
reganq, line 17: I don't think you should be calling the error view directly from the presenter. Rather, you should add a method to the viewmodel and use the observer design pattern

src/main/java/delete_post/use_case/DeletePostRequestModel.java
```diff
@@ -2,31 +2,23 @@
 
 import entities.CurrentUser;
 import entities.User;
-import entities.Post;
 
 public class DeletePostRequestModel {
-    private Post post;
+    int postId;
     private boolean isAdmin = CurrentUser.getIsAdmin();
     private User currentUser = CurrentUser.getCurrentUser();
```
reganq, line 9: I don't think the currentuser stuff needs to be stored here, as it can be accessed directly from the interactor

reganq (changes requested):
Looking good! I think we need to make a few changes though. 
- Firstly, the presenter should be calling the view model, which should be the only thing interacting with the view
- While on the topic of the interaction between the view and the viewmodel, I think you need to implement the observer design pattern. This can be seen in some other use cases
- You should create a deletepost view that just shows the delete button (implements jpanel) and is the one that calls the controller to start the use case
- I think you also need to remove the post from the user that created it
- We should be overwriting the csv file at the end, as this allows us to preserve ids of posts that have been deleted and can tell us that they've actually been deleted
- As always, documentation and tests are much needed!

src/main/java/delete_post/interface_adapters/DeletePostPresenter.java
```diff
@@ -12,10 +14,11 @@ public DeletePostPresenter(DeletePostViewModel viewModel){
...
     }
     @Override
-    public void prepareFailView(String error){
-        //error popup, no change to posts view or database;
+    public void prepareFailView(){
+        viewModel.updateViewModel(false, true);
     }
     public void prepareSuccessView(DeletePostResponseModel responseModel){
-        //update view posts view
+        viewModel.updateViewModel(true, true);
```
reganq, line 21: I think you need to change the second boolean to false, otherwise it will always cause the error message to show

reganq (changes requested):
Mostly looking good. I have a couple changes I think would help: most importantly we'll need to fix the gradle build and git ignore and things like that, as they will cause issues if pushed into main. Otherwise, I think there's a thing with the isTimer that I commented about. Lastly we need to fix the currentuser file, as that will cause merge issues

src/main/java/delete_post/use_case/DeletePostInteractor.java
```diff
@@ -22,7 +22,7 @@ public void delete(DeletePostRequestModel requestModel){
             this.outputBoundary.prepareTimerView(responseModel);
         }
         else if (CurrentUser.getIsAdmin() ||
-                CurrentUser.getCurrentUser() == post.getUser()){
+                CurrentUser.getCurrentUser().getId() == post.getUser()){
```
reganq, line 25: i think instead we should have CurrentUser.getCurrentUser().equals(post.getUser), as then we're checking equality of the data types

reganq: Wait nevermind I realized the Post method returns an id

reganq (dismissed):
Okay looks good and we can figure out the gradle stuff in tutorial

src/main/java/delete_post/UI/DeleteView.java
```diff
@@ -0,0 +1,38 @@
...
+        deleteButton.addActionListener(this);
+    }
+
+    public void actionPerformed(ActionEvent e){
+        controller.delete(this.postId, false);
+    }
+
+    public void propertyChange(PropertyChangeEvent event){
+        JFrame statusFrame = new JFrame();
+        statusFrame.setVisible(true);
```
sushimon, line 29: Is there a use for this blank frame? JOptionPane creates a new window anyways.

src/main/java/delete_post/drivers/DeletePostDataAccess.java
```diff
@@ -1,25 +1,118 @@
...
     }
+
     public void deletePost(int postId){
 
+        List<String[]> postData = readFile(postPath);
+
+        for (String[] row: postData){
+            if (Integer.parseInt(row[0]) == postId){
+                for (int i=1; i<9; i++){
+                    row[i] = null;
```
JohannLeidenfrost, line 57: Does this work when you try to write to the csv file? It seems like there may be an error when an entire row is just set to null. If it works, is there a blank line in the csv file?

JohannLeidenfrost: I would still suggest you copy all the posts into a new List<String[]> except the one that is to be deleted.

src/main/java/delete_post/interface_adapters/DeletePostPresenter.java
```diff
@@ -2,7 +2,9 @@
 
 import delete_post.use_case.DeletePostOutputBoundary;
 import delete_post.use_case.DeletePostResponseModel;
-import entities.User;
+import delete_post.UI.DeleteView;
+
+import javax.swing.JFrame;
```
sushimon, line 7: Unused import?

src/main/java/delete_post/drivers/DeletePostDataAccess.java
lisan-y: the id column is still there, so its not an empty row. is that how we decided we wanted it?

lisan-y: rather than setting the entire row to null, that for loop is setting columns 1-8 to null, and leaving the id column 0 as is

JohannLeidenfrost: I think that was how we wanted it originally but we then decided it would be better if we just delete the entire entry since we don't need it. If the id still exists, this may also be a problem for viewpost since it can't display a post where most of the entries are null. So could you please change it to delete the entire row?

src/test/java/entities/PostTest.java
```diff
@@ -30,48 +30,56 @@ public void tearDown() {
...
         newPost.removeFavouritedUser(user);
+
         assertEquals(0, newPost.getFavouritedUsers().size());
     }
 
     @Test
     public void setTags() {
         LocalDate deadline = LocalDate.of(2018, 12, 31);
-        Post newPost = new Post(poster, title, mainDesc, tags, collaborators, deadline, 0);
+        Post newPost = new Post(poster.getId(), title, mainDesc, tags, collaborators, deadline, 0);
```
sushimon, line 82: Isn't `poster` just an int as declared on line 15? There's no need to call `getId`.

sushimon (dismissed):

src/test/java/entities/PostTest.java
reganq: I think I resolved this in the merge conflicts, not exactly sure how it works though

lisan-y: ah my bad i was looking at an old version where poster was user

src/main/java/delete_post/UI/DeleteView.java
sushimon: Nevermind, you need a frame for the JOptionPane

src/main/java/delete_post/drivers/DeletePostDataAccess.java
```diff
@@ -1,25 +1,118 @@
...
+                StringBuilder newList = new StringBuilder();
+                String[] oldList = row[listCol].split(" ");
+                for (String id : oldList){
+                    if (Integer.parseInt(id) != postId) {
+                        if (newList.length() != 0){
+                            newList.append(" ");
+                        }
+                        newList.append(id);
+                    }
+                }
```
JohannLeidenfrost, line 111: I think when you run this, it will add a space at the end of the newList. So then you'll have a StringBuilder like "1 2 3 ", which may be a problem for other use cases. Can you remove the last space in the StringBuilder? Except if newList is empty or just of length 1 so if it is like "" or "1".

src/main/java/delete_post/drivers/DeletePostDataAccess.java
```diff
@@ -1,25 +1,118 @@
...
+        try {
+            FileWriter newFile = new FileWriter(path);
+            CSVWriter writer = new CSVWriter(newFile);
+            writer.writeAll(data);
+            writer.close();
+        } catch (IOException e) {
+            System.out.println("Error during writing file.");
+        }
+    }
+
```
JohannLeidenfrost, line 94: If there is time, you could also add a method for removing the comments to the post that will be deleted.

JohannLeidenfrost (changes requested):
Please address the comments in DeletePostDataAccess. The rest looks good!

reganq (approved):
Looks good! I think that we should merge this and then create a new issue to integrate it with the view

src/main/java/delete_post/drivers/DeletePostDataAccess.java
```diff
@@ -1,25 +1,127 @@
...
+    @Override
+    public Post getPost(int postId){
+        List<String[]> postData = readFile(postPath);
+        for (String[] row : postData) {
+            if (Integer.parseInt(row[0]) == postId) {
+                return postReader.readPost(row);
+            }
+        }
+        System.out.println("Could not find post in database");
+        return null;
```
JohannLeidenfrost, line 44: this could be a problem as it will cause errors in the program if you can't find the post for some reason. If there is time, you could implement an error popup that is displayed when the post cannot be found. ViewPost has exactly the same thing.

JohannLeidenfrost (changes requested):
Code looks good. However, there are a bunch of GitHub autograding failures partly because there are some duplicate classes in your branch. There is a user_case folder that contains the same classes as the use_case folder and there is a database folder that contains different db files than the Database folder. Please delete those files that are not supposed to be in your branch and check for errors again after.

JohannLeidenfrost (changes requested):
Could you comment your tests out or remove them from the PR since they're incomplete and will cause errors in others' local repos?

JohannLeidenfrost (approved):
Lgtm!

JohannLeidenfrost (approved):
Lgtm!

reganq (approved):
lgtm

src/test/java/delete_post/DeletePostTest.java
```diff
@@ -0,0 +1,100 @@
...
+        userBody.add(user0);
+        userBody.add(user1);
+        userBody.add(user2);
+
+        createTestFile(postPath, postHeader, postBody);
+        createTestFile(userPath, userHeader, userBody);
+    }
+
+    @Test
+    public void deletePostAdminSuccess(){
```
sushimon, line 59: Empty tests? Is this something we're adding later?

sushimon (approved):
No glaring issues, looks good to merge. General comments/concerns below.

Comments/concerns:
- Needs more documentation
- The tests in `DeletePostTest` are empty

PR #64, opened 2022-11-30
<High Priority> ViewPost 

izumidayo:
- everything but the tests are implemented (tested the UI elements with a dummy main file which is not included in this PR)
- the buttons on the UI at the moment do not react to anything. we need to work to merge the other usecase views with this one. (the main issue si that a lot of people need to have their controllers passed to them to initialize the view which we need to make the Main file to accomplish)
- the UI has all the functions it's supposed to have but I am struggling with Java Swing and I cannot control where the panels place themselves on another panel. I would appreciate it if some one could take a jab at fixing the aesthetics later.


src/main/java/view_post/ui/PostListView.java
```diff
@@ -1,4 +1,103 @@
...
+        }
+    }
+
+    /**
+     * When an item is selected in the scrollable list, the ViewPost use case is triggered and displays the
+     * post in the larger panel
+     * @param e the event that characterizes the change.
+     */
+    @Override
+    public void valueChanged(ListSelectionEvent e) {
```
JohannLeidenfrost, line 98: I don't really understand how this one works. Is this method called in the PostListView?

izumidayo: To my understanding, this is kind of like an action listener for the JList. When a post on the list is selected (no submit button needed), the controller is called to display the post on another panel.

JohannLeidenfrost (dismissed):
This looks great!

src/main/java/view_post/ui/PostListView.java
JohannLeidenfrost: Got it, thank you for explaining!

izumidayo: So in other words, the method is never explicitly called

src/main/java/view_post/interface_adapters/ViewPostPresenter.java
```diff
@@ -1,4 +1,31 @@
...
+        String postBody = responseModel.getPostBody();
+        String postTags = responseModel.getPostTags();
+        ArrayList<Integer> postReplies = responseModel.getPostReplies();
+        String deadline = responseModel.getDeadline();
+        String creationDate = responseModel.getCreationDate();
+        String collaborators = responseModel.getCollaborators();
+        int postID = responseModel.getPostID();
+        String title = responseModel.getTitle();
+
+        ViewPostOutputData outputData = new ViewPostOutputData(postEmail, postBody, postTags, postReplies, deadline,
```
JohannLeidenfrost, line 26: The GitHub auto grading is not finding the ViewPostOutputData class and I can't find it in your branch either. Did you forget to upload it or is it a work in progress?

izumidayo: GOOD CATCH it wasn't added to Git for some reason. I'm gonna push right now.

JohannLeidenfrost: Great!

JohannLeidenfrost (dismissed):
Looks good!

src/main/java/view_post/drivers/ViewPostGateway.java
```diff
@@ -1,14 +1,123 @@
...
+    /**
+     * Retrieves the post being favourited/unfavourited as a Post object from the database
+     *
+     * @param postid the id of the post being favourited/unfavourited
+     * @return a Post
+     */
+    private String[] getPostData(int postid){
+        try {
+            // finding the post in the database
+            List<String[]> allPosts = readAllLines(Paths.get(partialPath+"posts.csv"));
```
JohannLeidenfrost, line 90: Skip the first line here since the first line is just headers!

src/main/java/view_post/drivers/ViewPostGateway.java
```diff
@@ -1,14 +1,123 @@
...
     }
+    /**
+     * Retrieves email of a User from the database
+     * @param userid user's ID
+     * @return user's email
+     */
+    private String getUserEmail(int userid){
+        try {
+            // finding and retrieving the current user's data
+            List<String[]> allUsers = readAllLines(Paths.get(partialPath+"users.csv"));
```
JohannLeidenfrost, line 62: Skip the first line here. There will be an numberFormatException thrown here because it will try to do Integer.parseInt() on the header.

src/main/java/view_post/use_case/ViewPostResponseModel.java
```diff
@@ -19,12 +25,25 @@ public class ViewPostResponseModel {
...
     public ViewPostResponseModel(String posterEmail, String postBody, String[] postTags, String[] postReplies,
-                                 String deadline) {
+                                 String deadline, String creationDate, String collaborators, int postID, String title) {
         this.posterEmail = posterEmail;
         this.postBody = postBody;
-        this.postTags = postTags;
-        this.postReplies = postReplies;
+        String tags = "";
+        for (String tag : postTags){
+            tags.concat(", ");
```
JohannLeidenfrost, line 33: this needs to be tags = tags.concat(", ") otherwise it does not assign the concatenation to anything!

JohannLeidenfrost, line 33: this needs to be tags = tags.concat(", ") otherwise it does not assign the concatenation to anything!

JohannLeidenfrost, line 33: Also, if you add the comma first, then the tags will start with a comma.

src/main/java/view_post/use_case/ViewPostResponseModel.java
```diff
@@ -19,12 +25,25 @@ public class ViewPostResponseModel {
...
+            tags.concat(tag);
+        }
+        this.postTags = tags;
```
JohannLeidenfrost, line 36: unless you want tags to end with ", ", you need to remove the last two characters of the String.

src/main/java/view_post/drivers/ViewPostGateway.java
```diff
@@ -1,14 +1,125 @@
...
+        }catch (CsvException e2){
+            System.out.println("Error occurred while reading file");
+        }
+        return null;
+    }
+
+    /**
+     * Retrieves the post being favourited/unfavourited as a Post object from the database
+     *
+     * @param postid the id of the post being favourited/unfavourited
```
reganq, line 85: nitpick but i think this may be left over from a different use case

src/main/java/view_post/interface_adapters/ViewPostOutputData.java
```diff
@@ -0,0 +1,65 @@
...
+    private final String postBody;
+    private final String postTags;
+    private final ArrayList<Integer> postReplies;
+    private final String deadline;
+    private final String creationDate;
+    private final String collaborators;
+    private final int postID;
+    private final String title;
+
+    public ViewPostOutputData(String posterEmail, String postBody, String postTags, ArrayList<Integer> postReplies,
```
reganq, line 16: Note: we could make this smaller with a hashmap or something - super not pressing tho

src/main/java/view_post/interface_adapters/ViewPostViewModel.java
```diff
@@ -1,4 +1,21 @@
...
+    private final PropertyChangeSupport support;
+    private final ViewPostView view;
+
+    public ViewPostViewModel(ViewPostView view){
+        this.support = new PropertyChangeSupport(this);
+        this.view = view;
+    }
+
+    public void updateView(ViewPostOutputData data){
+        view.setData(data);
```
reganq, line 17: We don't need this function, as far as I know. It is questionable wrt CA, but we can actually get the data from the property change event

src/main/java/view_post/ui/ViewPostView.java
```diff
@@ -1,4 +1,188 @@
...
+    }
+
+    public void setData(ViewPostOutputData data){
+        this.data = data;
+    }
+
+    @Override
+    public void propertyChange(PropertyChangeEvent event){
+        try {
+            this.displayPost(this.data);
```
reganq, line 33: As I referenced above, we can populate this.data with event.getNewValue()

reganq (approved):
Wow, this is really great!! good job guys!

src/main/java/view_post/use_case/ViewPostResponseModel.java
```diff
@@ -19,12 +25,26 @@ public class ViewPostResponseModel {
      * @param deadline      The string representation of the deadline of this post.
      */
     public ViewPostResponseModel(String posterEmail, String postBody, String[] postTags, String[] postReplies,
-                                 String deadline) {
+                                 String deadline, String creationDate, String collaborators, int postID, String title) {
         this.posterEmail = posterEmail;
         this.postBody = postBody;
-        this.postTags = postTags;
-        this.postReplies = postReplies;
+        String tags = "";
```
sushimon, line 31: Rather than doing all this, since the parameter `postTags` is an array of strings, you can just use `String.join`. Less likely to be buggy.

src/main/java/view_post/use_case/ViewPostResponseModel.java
```diff
@@ -19,12 +25,26 @@ public class ViewPostResponseModel {
...
+            tags = tags.concat(tag);
+            tags = tags.concat(", ");
+        }
+        tags = tags.substring(0, tags.length() - 2);
+        this.postTags = tags;
+        ArrayList replies = new ArrayList<>();
+        for (String reply : postReplies){
+            replies.add(Integer.valueOf(reply));
+        }
+        this.postReplies = replies;
```
sushimon, line 42: Can just initialize `postReplies` as an empty array list and then do the loop after. Removes the need for line 38. Also, as it is, does this not highlight line 38 for using a raw type? It doesn't look like you specified a type for `replies`.

src/main/java/view_post/use_case/ViewPostResponseModel.java
```diff
@@ -44,14 +64,14 @@ public String getPostBody() {
...
-    public String[] getPostTags() {
+    public String getPostTags() {
         return postTags;
     }
 
     /**
      * Return a list of replies made to this post.
      */
-    public String[] getPostReplies() {
+    public ArrayList<Integer> getPostReplies() {
```
sushimon, line 74: Also, why is `postReplies` an ArrayList? If we want to add to it, then we should add a method to add elements to the list rather than calling this getter and then directly calling the add method of ArrayList.

sushimon (changes requested):
Comments/concerns:
- Is the `posts.csv` file supposed to be deleted? 
- Add more documentation to classes and stuff.
- Make sure variable names and such follow standard naming conventions. I saw multiple variables that are in all lowercase despite being multiple words.

src/main/java/view_post/drivers/ViewPostGateway.java
```diff
@@ -1,14 +1,125 @@
...
+import java.io.Reader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+
 public class ViewPostGateway implements ViewPostDsGateway {
+    // partialpath is a String that contains the partial directory that leads to all csv files in the program
+    // it is isolated from the file name so that in case the file is moved, the code is still compatible
+    private final String partialPath;
```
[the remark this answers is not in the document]
izumidayo: I'm going to leave this as is for now because the other use cases are taking the same approach

src/main/java/view_post/drivers/ViewPostGateway.java
```diff
@@ -1,14 +1,125 @@
...
+            List<String[]> allPosts = readAllLines(Paths.get(partialPath+"posts.csv"));
+            allPosts.subList(0,1).clear();
+            String[] postData = new String[10];
+            for (String[] post : allPosts) {
+                try {
+                    int id = Integer.parseInt(post[0]);
+                    if (id == postid) {
+                        postData = post;
+                    }
+                } catch (NumberFormatException ex) {
```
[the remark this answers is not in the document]
izumidayo: I added a comment in my code describing the purpose of catching and ignoring this exception. please let me know if you think this is okay or if there is a better way to deal with it

src/main/java/view_post/interface_adapters/ViewPostViewModel.java
```diff
@@ -1,4 +1,21 @@
...
+        support.firePropertyChange("show post", null, data);
```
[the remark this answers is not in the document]
izumidayo: I was thinking that would be okay becuase it would basically just be refreshing.

src/main/java/view_post/interface_adapters/ViewPostOutputData.java
izumidayo: Might work on this later...

src/main/java/view_post/drivers/ViewPostGateway.java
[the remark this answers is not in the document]
izumidayo: Added a print line statement just so that there is something there

reganq (approved):
Looks good to me!

Thehashhobo:
could use some tests

BerryRahul (approved):
lgtm!

Thehashhobo (approved):
lgtm

PR #65, opened 2022-11-30
4 feature 9 make a new post

JohannLeidenfrost:
I added documentation and tests. Everything is tested except for the makePostView.


reganq (dismissed):
Looks good!

src/main/java/make_post/interface_adapters/MakePostController.java
```diff
@@ -20,23 +20,29 @@ public MakePostController(MakePostInputBoundary interactor){
         this.interactor = interactor;
     }
 
+    /**
+     * passes the input made by the user to make the post to the interactor.
+     * @param postBody the attributes of the post input by the user.
+     */
     public void passToMakePostInteractor(Map<String, Object> postBody){
```
sushimon, line 27: Nitpick but you could just name this something like `execute` or `makePost` or something. As it is, it's kinda too technical.

src/test/java/filter_post/FilterPostTest.java
```diff
@@ -191,7 +191,7 @@ public void testDsGatewayCatchIOException() {
     @Test
     public void testDsGatewayCatchCsvException() {
         // The csv file contains a row that does not have the same number of columns as the header row.
-        String path = "src/test/java/filter_post/error.csv";
```
sushimon, line 194: This was supposed to be intentional. The `error.csv` file was designed so that a `CsvException would be thrown. Although this test still passes, it does not test the specified behaviour.

sushimon (changes requested):
Comments/concerns:
- Did we discuss why why want the path to just contain part of the directory? If they move the file or anything we shouldn't be responsible for what happens then imo if we just state that the csv files must be in a certain directory.
- Some variables aren't following naming conventions. Some variables are entirely lowercase despite being multiple words
- Revert the changes made to `FilterPostTest.java`. I believe they still pass but it no longer tests the intended behaviour of that unit test. See the comments on the file for more details.

src/test/java/filter_post/FilterPostTest.java
JohannLeidenfrost: So sorry I think I meant to change that in my tests but then had your test open and and accidentally changed that there!

src/main/java/make_post/interface_adapters/MakePostController.java
JohannLeidenfrost: What do you mean by the comment about the path containing part of the directory?

src/main/java/make_post/use_case/MakePostInteractor.java
```diff
@@ -12,9 +12,13 @@
 public class MakePostInteractor implements MakePostInputBoundary {
     private final MakePostDsGateway dataAccess;
     private final MakePostOutputBoundary presenter;
-
     private final PostFactory postFactory = new PostFactory();
```
reganq, line 15: The only thing is that the post probably should be passed in the constructor, but thats not a huge issue and I can add this as I'm working on putting the different iterations of post and user factory together into one

reganq (approved):
This looks good to me!

izumidayo (approved):
Looks good to me!

BerryRahul (approved):
looks good to merge

sushimon (approved):

PR #67, opened 2022-12-01
Main File + Main Frame

reganq:
Current use case for the main frame and the main method. Stopped before setting up the observers because I got tired. If everyone wants to take a look at this and lmk what changes need to be made to initialize your use case + what observers need to be set up that would be a life saver.

### DIFF
--- a/src/main/java/main.java
+++ b/src/main/java/main.java
@@ -1,0 +1,137 @@
+import delete_post.drivers.DeletePostDataAccess;
+import delete_post.interface_adapters.DeletePostController;
+import delete_post.interface_adapters.DeletePostPresenter;
+import delete_post.interface_adapters.DeletePostViewModel;
+import delete_post.use_case.DeletePostInteractor;
+import filter_post.drivers.FilterPostDataAccess;
+import filter_post.interface_adapters.FilterPostController;
+import filter_post.interface_adapters.FilterPostPresenter;
+import filter_post.interface_adapters.FilterPostViewModel;
+import filter_post.ui.FilterPostBarView;
+import filter_post.use_case.FilterPostInteractor;
+import make_post.drivers.MakePostDatabaseAccess;
+import make_post.interface_adapters.MakePostController;
+import make_post.interface_adapters.MakePostPresenter;
+import make_post.interface_adapters.MakePostViewModel;
+import make_post.ui.MakePostView;
+import make_post.use_case.MakePostInteractor;
+import make_post.use_case.PostFactory;
+import sign_up.drivers.SignUpDatabaseAccess;
+import sign_up.interface_adapters.SignUpController;
+import sign_up.interface_adapters.SignUpPresenter;
+import sign_up.interface_adapters.SignUpViewModel;
+import sign_up.ui.MasterLandingView;
+import sign_up.ui.SignUpView;
+import sign_up.use_case.SignUpInteractor;
+import sign_up.use_case.UserFactory;
+import view_post.drivers.ViewPostGateway;
+import view_post.interface_adapters.ViewPostController;
+import view_post.interface_adapters.ViewPostPresenter;
+import view_post.interface_adapters.ViewPostViewModel;
+import view_post.ui.MainFrame;
+import view_post.ui.PostListView;
+import view_post.ui.ViewPostView;
+import view_post.use_case.ViewPostInteractor;
+
+/*
+* The main class for the file, containing the main method which runs the program
+* */
+public class main {
+    static String adminPath = new String("");
+    static String numPostsCreatedFilePath = new String("");
+    static String numUsersCreatedFilePath = new String("");
+    static String numCommentsCreatedFilePath = new String("");
+    static String postsFilePath = new String("");
+    static String usersFilePath = new String("");
+    static String commentsFilePath = new String("");
+    static String generalPath = new String("");
+
+    public static void main(String[] args) {
+        // initialize stuff for delete_post
+        DeletePostViewModel deletePostViewModel = new DeletePostViewModel();
+        DeletePostPresenter deletePostPresenter = new DeletePostPresenter(deletePostViewModel);
+        DeletePostDataAccess deletePostDataAccess = new DeletePostDataAccess(generalPath);
+        DeletePostInteractor deletePostInteractor = new DeletePostInteractor(deletePostPresenter, deletePostDataAccess);
+        DeletePostController deletePostController = new DeletePostController(deletePostInteractor);
+
+        // initialize stuff for filter_post
+        FilterPostViewModel filterPostViewModel = new FilterPostViewModel(new String[0], new int[0], new String[0]);
+        FilterPostPresenter filterPostPresenter = new FilterPostPresenter(filterPostViewModel);
+        FilterPostDataAccess filterPostDataAccess = new FilterPostDataAccess(postsFilePath);
+        FilterPostInteractor filterPostInteractor = new FilterPostInteractor(filterPostDataAccess, filterPostPresenter);
+        FilterPostController filterPostController = new FilterPostController(filterPostInteractor);
+        FilterPostBarView filterPostBarView = new FilterPostBarView(new String[0], filterPostController);
+
+        // initialize stuff for make_post
+        MakePostViewModel makePostViewModel = new MakePostViewModel();
+        MakePostPresenter makePostPresenter = new MakePostPresenter(makePostViewModel);
+        MakePostDatabaseAccess makePostDatabaseAccess = new MakePostDatabaseAccess(generalPath);
+        MakePostInteractor makePostInteractor = new MakePostInteractor(makePostDatabaseAccess, makePostPresenter);
+        MakePostController makePostController = new MakePostController(makePostInteractor);
+        MakePostView makePostView = new MakePostView(new String[0], makePostController);
+
+        // initialize stuff for sign_up
+        SignUpViewModel signUpViewModel = new SignUpViewModel();
+        SignUpPresenter signUpPresenter = new SignUpPresenter(signUpViewModel);
+        SignUpDatabaseAccess signUpDatabaseAccess = new SignUpDatabaseAccess(usersFilePath, adminPath, numUsersCreatedFilePath);
+        SignUpInteractor signUpInteractor = new SignUpInteractor(signUpDatabaseAccess, signUpPresenter);
+        SignUpController signUpController = new SignUpController(signUpInteractor);
+        SignUpView signUpView = new SignUpView(signUpController);
+
+        // initialize stuff for make_comment
+        MakeCommentViewModel makeCommentViewModel = new MakeCommentViewModel();
+        MakeCommentPresenter makeCommentPresenter = new MakeCommentPresenter(makeCommentViewModel);
+        MakeCommentDatabaseAccess makeCommentDatabaseAccess = new MakeCommentDatabaseAccess(numCommentsCreatedFilePath, commentsFilePath);
+        MakeCommentInteractor makeCommentInteractor = new MakeCommentInteractor(makeCommentDatabaseAccess, makeCommentPresenter);
+        MakeCommentController makeCommentController = new MakeCommentController(makeCommentInteractor);
+        MakeCommentView makeCommentView = new MakeCommentView(makeCommentController);
+
+        // initialize stuff for timer
+        PostFactory postFactory = new PostFactory();
+        TimerDataAccess timerDataAccess = new TimerDataAccess(postsFilePath, postFactory);
+        TimerInteractor timerInteractor = new TimerInteractor(timerDataAccess, deletePostInteractor);
+        TimerController timerController = new TimerController(timerInteractor);
+
+        // initialize stuff for log in
+        LogInViewModel logInViewModel = new LogInViewModel();
+        LogInPresenter logInPresenter = new LogInPresenter(logInViewModel);
+        LogInDatabaseAccess logInDatabaseAccess = new LogInDatabaseAccess(usersFilePath);
+        LogInInteractor logInInteractor = new LogInInteractor(logInDatabaseAccess, logInPresenter);
+        LogInController logInController = new LogInController(logInInteractor);
+        LogInView logInView = new LogInView(logInController);
+
+        // initialize stuff for log out
+        LogOutViewModel logOutViewModel = new LogOutViewModel();
+        LogOutPresenter logOutPresenter = new LogOutPresenter(logOutViewModel);
+        LogOutInteractor logOutInteractor = new LogOutInteractor(logOutPresenter);
+        LogOutController logOutController = new LogOutController(logOutInteractor);
+        LogOutView logOutView = new LogOutView(logOutController);
+
+        // initialize stuff for favourite_post
+        UserFactory userFactory = new UserFactory();
+        FavourtiteViewModel favourtiteViewModel = new FavouriteViewModel();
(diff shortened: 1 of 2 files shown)


src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+
+        // initialize stuff for log out
+        LogOutViewModel logOutViewModel = new LogOutViewModel();
+        LogOutPresenter logOutPresenter = new LogOutPresenter(logOutViewModel);
+        LogOutInteractor logOutInteractor = new LogOutInteractor(logOutPresenter);
+        LogOutController logOutController = new LogOutController(logOutInteractor);
+        LogOutView logOutView = new LogOutView(logOutController);
+
+        // initialize stuff for favourite_post
+        UserFactory userFactory = new UserFactory();
```
izumidayo, line 111: Please add the code below!

PostFactory postFactory = new PostFactory();

src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+        FavourtiteViewModel favourtiteViewModel = new FavouriteViewModel();
+        FavouritePresenter favouritePresenter = new FavouritePresenter(favourtiteViewModel);
+        DataAccess dataAccess = new DataAccess(postFactory, userFactory, generalPath);
+        FavouriteInteractor favouriteInteractor = new FavouriteInteractor(dataAccess, favouritePresenter);
+        FavouriteController favouriteController = new FavouriteController(favouriteInteractor);
+        FavouriteView favouriteView = new FavouriteView(favouriteController);
```
izumidayo, line 117: Please remove this code as the view will be initialized in the main view with every refresh

src/main/java/main.java
izumidayo: unless we merge the different PostFactories into 1

izumidayo (changes requested):
Thank you for making this! I left a few comments for changes that need to be made for the favourite post section

src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+        // initialize stuff for filter_post
+        FilterPostViewModel filterPostViewModel = new FilterPostViewModel(new String[0], new int[0], new String[0]);
+        FilterPostPresenter filterPostPresenter = new FilterPostPresenter(filterPostViewModel);
+        FilterPostDataAccess filterPostDataAccess = new FilterPostDataAccess(postsFilePath);
+        FilterPostInteractor filterPostInteractor = new FilterPostInteractor(filterPostDataAccess, filterPostPresenter);
+        FilterPostController filterPostController = new FilterPostController(filterPostInteractor);
+        FilterPostBarView filterPostBarView = new FilterPostBarView(new String[0], filterPostController);
+
+        // initialize stuff for make_post
+        MakePostViewModel makePostViewModel = new MakePostViewModel();
```
JohannLeidenfrost, line 66: This is all that needs to be initialised for Make Post. Thanks again!

src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+        // initialize stuff for delete_post
+        DeletePostViewModel deletePostViewModel = new DeletePostViewModel();
+        DeletePostPresenter deletePostPresenter = new DeletePostPresenter(deletePostViewModel);
+        DeletePostDataAccess deletePostDataAccess = new DeletePostDataAccess(generalPath);
+        DeletePostInteractor deletePostInteractor = new DeletePostInteractor(deletePostPresenter, deletePostDataAccess);
+        DeletePostController deletePostController = new DeletePostController(deletePostInteractor);
+
+        // initialize stuff for filter_post
+        FilterPostViewModel filterPostViewModel = new FilterPostViewModel(new String[0], new int[0], new String[0]);
+        FilterPostPresenter filterPostPresenter = new FilterPostPresenter(filterPostViewModel);
```
sushimon, line 59: The type of this should be `FilterPostOutputBoundary`

src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+        FilterPostDataAccess filterPostDataAccess = new FilterPostDataAccess(postsFilePath);
```
sushimon, line 60: The type of this should be `FilterPostDsGateway`

src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+        FilterPostInteractor filterPostInteractor = new FilterPostInteractor(filterPostDataAccess, filterPostPresenter);
```
sushimon, line 61: The type of this should be `FilterPostInputBoundary`

src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+        FilterPostController filterPostController = new FilterPostController(filterPostInteractor);
+        FilterPostBarView filterPostBarView = new FilterPostBarView(new String[0], filterPostController);
```
sushimon, line 63: We need to come up with a preset list of tags so we can pass that in here as the first argument

src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+    public static void main(String[] args) {
+        // initialize stuff for delete_post
+        DeletePostViewModel deletePostViewModel = new DeletePostViewModel();
+        DeletePostPresenter deletePostPresenter = new DeletePostPresenter(deletePostViewModel);
+        DeletePostDataAccess deletePostDataAccess = new DeletePostDataAccess(generalPath);
+        DeletePostInteractor deletePostInteractor = new DeletePostInteractor(deletePostPresenter, deletePostDataAccess);
+        DeletePostController deletePostController = new DeletePostController(deletePostInteractor);
+
+        // initialize stuff for filter_post
+        FilterPostViewModel filterPostViewModel = new FilterPostViewModel(new String[0], new int[0], new String[0]);
```
sushimon, line 58: Need to add `PostListView` from `src/main/java/view_post/ui` as an observer to the view model when we actually connect things

src/main/java/main.java
```diff
@@ -0,0 +1,137 @@
...
+import view_post.interface_adapters.ViewPostViewModel;
+import view_post.ui.MainFrame;
+import view_post.ui.PostListView;
+import view_post.ui.ViewPostView;
+import view_post.use_case.ViewPostInteractor;
+
+/*
+* The main class for the file, containing the main method which runs the program
+* */
+public class main {
```
sushimon, line 39: Improper naming convention for classes. Needs to be uppercase but refer to general review comments for suggested changes.

sushimon (changes requested):
General comments/concerns:
- Remember that in Clean Architecture, we shouldn't be referring to certain implementing classes directly when declaring data types. Replace them with the interfaces instead. For example, when initializing an interactor, it should be of type `<use_case>InputBoundary`. Just taking a brief look, you need to change this for all of the `<use_case>DataAccess`, `<use_case>Interactor`, and `<use_case>Presenter` classes. We are using `<use_case>DsGateway`, `<use_case>InputBoundary`, and `<use_case>OutputBoundary` as interface names for the respective implementing classes listed previously. 
- For the filter post use case, the only setup that is needed to be done is adding `PostListView` as an observer of the `FilterPostViewModel`.
- We need to come up with a preset string array of tags. Also once this is done, it needs to be passed in as an argument to `FilterPostBarView`.
- Maybe we should just rename the class to `UofTeams` instead of `main`. 
- Is there a reason that `MainFrame.java` is in the `view_post/ui` directory?

src/main/java/UofTeams.java
```diff
@@ -0,0 +1,187 @@
...
+        HeaderView headerView = new HeaderView(generalPath);
+
+
+        // initialize main view
+        MasterLandingView masterLandingView = new MasterLandingView(signUpView, logInView);
+        MainPostView mainPostView = new MainPostView(viewPostView, postListView, headerView);
+        MainFrame mainFrame = new MainFrame(masterLandingView, mainPostView);
+
+        // Setting up observers
+        // filterPostViewModel.addObserver(postListView);
```
sushimon, line 171: Why is this commented out?

src/main/java/filter_post/use_case/FilterPostInteractor.java
```diff
@@ -31,7 +31,7 @@ public void filterPosts(FilterPostRequestModel filters) {
 
         for (String[] postInfo : posts) {
             String postTags = postInfo[4];
-            if (containsAllWords(postTags, filterTags)) {
+            if (filterTags.length == 0 || containsAllWords(postTags, filterTags)) {
```
sushimon, line 34: Redundant condition. `containsAllWords` already returns true if `filterTags` is empty.

sushimon (changes requested):

src/main/java/UofTeams.java
reganq: Not sure actually, I'll change that

src/main/java/filter_post/use_case/FilterPostInteractor.java
reganq: Got it, will update

src/main/java/UofTeams.java
```diff
@@ -0,0 +1,187 @@
...
+        makePostViewModel.addObserver(makePostView);
+
```
izumidayo, line 173: Do you think it would be okay if I did this for FavouriteView in the MainPostView in the view_post ui?
The Favourite View gets initialized every time the view gets refreshed so I can't handle it here. 
HOWEVER if that is a bad idea, I can also create a setter for the FavouriteView that will pass new post IDs to the view every time ViewPostView is refreshed

src/main/java/UofTeams.java
```diff
@@ -0,0 +1,233 @@
...
+        DeleteView deleteView = new DeleteView(deletePostController);
+
+        // initialize stuff for filter_post
+        FilterPostViewModel filterPostViewModel = new FilterPostViewModel(new String[0], new int[0], new String[0]);
+        FilterPostOutputBoundary filterPostPresenter = new FilterPostPresenter(filterPostViewModel);
+        FilterPostDsGateway filterPostDataAccess = new FilterPostDataAccess(postsFilePath);
+        FilterPostInputBoundary filterPostInteractor = new FilterPostInteractor(filterPostDataAccess, filterPostPresenter);
+        FilterPostController filterPostController = new FilterPostController(filterPostInteractor);
+        // Requires array of preset tags
+        FilterPostBarView filterPostBarView = new FilterPostBarView(new String[]{"sports", "quantum", "math"}, filterPostController);
```
sushimon: Just leaving a comment here as a reference: we need to come up with a list of tags and update that here.

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -64,17 +64,24 @@ public boolean checkPasswordMatches(String email, String pass){
     }
 
     @Override
-    public ArrayList<String> getUser(boolean success, String email, String pass){
-        ArrayList<String> userInfo = new ArrayList<>();
+    public String[] getUser(boolean success, String email, String pass){
```
BerryRahul, line 67: the boolean success is not needed as the logininteractor already checks this, it can be removed.

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -64,17 +64,24 @@ public boolean checkPasswordMatches(String email, String pass){
...
+        String[] userInfo = new String[6];
         if (success){
```
BerryRahul, line 69: Can be removed as per last comment

src/main/java/log_in/drivers/LogInDatabaseAccess.java
```diff
@@ -64,17 +64,24 @@ public boolean checkPasswordMatches(String email, String pass){
...
-            userInfo.add(pass);
-            userInfo.add(adminValueString);
+
+            userInfo[0] = ids.get(emailIndex);
+            userInfo[1] = adminValueString;
+            userInfo[2] = email;
+            userInfo[3] = pass;
+            userInfo[4] = posts.get(emailIndex);
+            userInfo[5] = favs.get(emailIndex);
         } else {
```
BerryRahul, line 85: can be removed based on last comment

src/main/java/log_in/use_case/LogInDsGateway.java
```diff
@@ -13,7 +11,7 @@ public interface LogInDsGateway {
     boolean checkPasswordMatches(String email, String pass);
 
     // gets the user with the given email, password
-    ArrayList<String> getUser(boolean success, String email, String pass);
+    String[] getUser(boolean success, String email, String pass);
```
BerryRahul, line 14: boolean success can be removed

src/main/java/log_in/use_case/LogInInteractor.java
```diff
@@ -76,7 +69,8 @@ private LogInResponseModel logInCheck(LogInRequestModel requestModel) throws Use
             throw new UserException("Incorrect Password");
         }
         if (this.checkPassword(requestModel.getEmail(), requestModel.getPassword())){
-            setCurrentUser(requestModel.getEmail(), requestModel.getPassword());
+            User loggedIn = userFactory.readUser(access.getUser(true, requestModel.getEmail(), requestModel.getPassword()));
```
BerryRahul, line 72: the true param can be removed as per other reviews

src/main/java/make_comment/ui/MakeCommentView.java
```diff
@@ -43,11 +43,18 @@ public void setPostID(int postId){
...
      */
     @Override
+    @SuppressWarnings("all")
     public void actionPerformed(ActionEvent e) {
             JFrame frame = new JFrame("commentate");
             String commentBody = JOptionPane.showInputDialog(frame, "Please enter the Comment",
                     "commentate", JOptionPane.QUESTION_MESSAGE);
-            controller.passToInteractor(commentBody, this.postId);
+
+            if (commentBody == null){
```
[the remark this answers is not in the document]
Thehashhobo: i see

src/main/java/view_post/ui/PostListView.java
```diff
@@ -24,28 +25,27 @@ public class PostListView extends JPanel implements PropertyChangeListener, List
...
-        barView = filterBar;
         this.add(filterBar);
         this.titles = null;
         this.ids = null;
         this.controller = controller;
         this.list = null;
 
         // Default view when there are no posts to show
-        JLabel noPostsText = new JLabel("Press Search :)");
+        JLabel noPostsText = new JLabel("Press Search");
```
izumidayo, line 48: put it backkk

src/main/java/log_out/use_case/LogOutRequestModel.java
```diff
@@ -2,18 +2,12 @@
...
 
-    private final String logOutRequest;
-
 
     /**
      * Initialize a log_out Request
      * @param logOutRequest A string to indicate a logout request has been made
      */
     public LogOutRequestModel(String logOutRequest){
-        this.logOutRequest  = logOutRequest;
```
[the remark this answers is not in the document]
JohannLeidenfrost: Will leave this for now as it is being dealt with in another PR.

src/main/java/make_comment/ui/MakeCommentView.java
```diff
@@ -43,11 +43,18 @@ public void setPostID(int postId){
...
+
```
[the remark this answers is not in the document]
JohannLeidenfrost: Will leave this too as it is being dealt with somewhere else.

src/main/java/view_post/ui/PostListView.java
JohannLeidenfrost: Done :)

PR #73, opened 2022-12-04
Database + factory consolidation

reganq:
Added general factories and database access for the use cases in the main branch at the time (filter, log in, sign up, make post). Will add for others when the pull requests have been merged (favourite, comment, delete, timer)


sushimon (commented):
Comments/concerns:
- Might just be me being out of the loop, but what's the point of having the general database access file `DataAccess.java` if we already have the data access for each individual use case? IMO it just makes the code more coupled for no reason and I feel that it probably violates one of the SOLID principles.
- Still am unsure about having these general factories, etc. Seems like it may potentially violate either the single responsibility principle. E.g. if one use case were to potentially change how they want posts to be created, it would change the other use cases too.

src/main/java/drivers/DataAccess.java
```diff
@@ -0,0 +1,387 @@
...
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+ public class DataAccess implements FilterPostDsGateway, LogInDsGateway, MakePostDsGateway, SignUpDsGateway{
```
izumidayo, line 27: Please update to implement ViewPostDsGateway, FavouriteDsGateway!

izumidayo (commented):
Made a comment about needing to include Favourite and View Post use cases but I can actually do that in a new pull request after this gets merged,

reganq:
Edit: My commit should say ViewPostDsGateway I think, as opposed to ViewPostInteractor.
Things that can be added to this PR can be to get rid of duplicate code by turning it into one private method

sushimon (dismissed):

sushimon (approved):

PR #85, opened 2022-12-04
5 feature 1 view post

JohannLeidenfrost:
Fixed some tests and removed ViewPostView from constructor for ViewPostViewModel.

### DIFF
--- a/src/main/java/view_post/interface_adapters/ViewPostViewModel.java
+++ b/src/main/java/view_post/interface_adapters/ViewPostViewModel.java
@@ -1,7 +1,5 @@
 package view_post.interface_adapters;
 
-import view_post.ui.ViewPostView;
-
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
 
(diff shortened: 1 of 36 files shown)


izumidayo (approved):
Looks good!

reganq (approved):
I don't think the csv can stay in the test file, but everything else is ok!

src/main/java/view_post/interface_adapters/ViewPostViewModel.java
```diff
@@ -10,15 +8,13 @@
  */
 public class ViewPostViewModel {
     private final PropertyChangeSupport observable;
-    private final ViewPostView view;
 
     /**
      * Initializes ViewPostViewModel.
      * @param view A ViewPostView object
      */
-    public ViewPostViewModel(ViewPostView view){
```
sushimon, line 19: Maybe I'm just out of the loop, but why is the view being removed from this class?

reganq (approved):
lgtm

izumidayo (approved):
Looks good I think

PR #101, opened 2022-12-06
100 fix duplicate classes delete post

JohannLeidenfrost:


izumidayo (approved):
Yeah this is what I had in mind. Looks good !

PR #110, opened 2022-12-06
Changes to allow instantiation of MakeCommentView and FavouriteView in the main method

izumidayo:
Updated parameters need for the initialization of MakeCommentView and FavouriteView

- removed postId parameter from the constructor of MakeCommentView and FavouriteView so that they can be instantiated in the main method
- added a setPostId method in each of the views 
- added MakeCommentView and FavouriteView as parameters for the ViewPostView constructor
- updated the displayPost method in ViewPostView so that it displays the MakeCommentView and FavouriteView AND calls the postId setter
- updated tests for MakeCommentView and ViewPostView accordingly

### DIFF
--- a/src/main/java/favourite/ui/FavouriteView.java
+++ b/src/main/java/favourite/ui/FavouriteView.java
@@ -15,19 +15,29 @@ public class FavouriteView extends JPanel implements ActionListener, PropertyCha
     // the controllwe that gets triggered when a user interacts with the FavouriteView
     private final FavouriteController controller;
     // the id of the post being favourited/unfavourited
-    private final int postid;
+    private int postId;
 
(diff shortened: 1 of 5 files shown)


reganq (approved):
This looks good to me!

JohannLeidenfrost (approved):
Lgtm!

reganq (approved):
Approving!

sushimon (approved):
No glaring issues, looks good to merge.

Comments/concerns:
- Not specific to this but for anything related to the `MakeComment` use case, we may need to do some refactoring to rename the variables. Variable names like `MCC` can be replaced with something like `controller` so that the variables are in camel case.

PR #123, opened 2022-12-07
Error frames issue fixed.

JohannLeidenfrost:
You always had to click twice to close an error popup because an error frame was made visible together with a JOptionPane.


reganq (approved):
Looks good!

izumidayo (approved):
Looks good, nice catch

PR #170, opened 2022-12-08
Display Favourited Posts and My Posts

izumidayo:
- played around with filter post data access and interactor so that a user can search for their own posts and favourited posts
- work to be done would be updating filter post testing and condensing the code in the filter post interactor class so we dont have as much duplicate code


Thehashhobo (approved):
looks good, I was under the impression that picking out individual data from csvs should be done in the interactor but it shouldn't matter.

reganq (dismissed):
approving!

BerryRahul (dismissed):
looks good!

BerryRahul (approved):

reganq (approved):
approved

Thehashhobo (approved):
lgtm

PR #187, opened 2022-12-08
Add tests for filter post use case and general refactoring.

sushimon:
Changes:
- Added new test cases in `FilterPostTest` to test when the user filters for their own posts and when they filter for the posts they've favourited
- Fixing code smells by extracting duplicate code in `FilterPostInteractor` into helper methods


JohannLeidenfrost (approved):
Lgtm as long as it still runs properly!

BerryRahul (approved):
lgtm

PR #190, opened 2022-12-08
commit to address warnings

Thehashhobo:
Attempt to fix warnings in the project.


src/test/java/log_in/LogInTest.java
```diff
@@ -22,7 +22,7 @@ public class LogInTest {
     LogInInputBoundary interactor;
     LogInOutputBoundary presenter;
     LogInController controller;
-    LogInUserFactory userFactory;
+    FavouriteUserFactory userFactory;
```
reganq, line 25: We'll have to fix this so it works with the LogInUserFactory instead, as that is the object being used in the log in use case

reganq (changes requested):
Can you refactor the log in test so it uses the LogInUserFactory?

BerryRahul (approved):
good changes, looks great

izumidayo (approved):
Everything looks good! Thanks Jerry

